PR #20, opened 2022-11-28
Fix `cargo-generate` arguments and binary URL

michalvavra:
- Link to download `cargo-generate` binaries had incorrect URL
  - This caused an issue when `cargo-generate` was not installed in PATH (via `cargo install cargo-generate)`
- Arguments for `cargo-generate` 
  - Missing `generate` argument (it's also needed when invoking the command from PATH `~/.cargo/bin/cargo-generate generate`)
  - `--name test` should be send to  `["--name", "test"]`

### DIFF
--- a/src/run/new.rs
+++ b/src/run/new.rs
@@ -54,6 +54,7 @@ impl NewCommand {
         let args = self.to_args();
         let exe = cargo_generate_exe()?;
         let mut process = Command::new(exe)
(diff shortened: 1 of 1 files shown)


akesson:
Thanks for the PR. I'm deep into something else. I'll have a look at the latest Thursday.

akesson:
All good. Thanks for the help.

PR #25, opened 2022-12-03
use git dependency

ptillemans:
This is a temporary fix as it is probably not a good idea to track the main branch. But I assume it will be fixed shortly? In any case this makes cargo-leptos work on my machine.


akesson:
Thanks a lot!

PR #44, opened 2023-01-11
Fix for single-package setup not triggering rebuild on edits

Indrazar:
In src/config/bin_package.rs and in src/config/lib_package.rs
Calling `rel_dir.join("src")` was appending:
  `.\\src` on Windows
  `./src` on Linux

The rest of the parsed folders are formatted without the leading ./ or .\\\\ leading to nothing ever being a "watched" item.

Changing the join to match how other directories are parsed gives:
`rel_dir.join("src").unbase(&rel_dir).expect()`

### DIFF
--- a/src/config/bin_package.rs
+++ b/src/config/bin_package.rs
@@ -85,7 +85,11 @@ impl BinPackage {
         };
 
         let mut src_paths = metadata.src_path_dependencies(&package.id);
-        src_paths.push(rel_dir.join("src").unbase(&rel_dir).expect("rel_dir should contain itself, it did not"));
+        if rel_dir == "." {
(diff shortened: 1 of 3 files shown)


Indrazar:
This does break the tests, I'll work on that now.
The following tests correctly fail:
config::tests::test_workspace_in_subdir_project2
config::tests::test_workspace
config::tests::test_workspace_project1
config::tests::test_workspace_project2

The following test incorrectly(?) fails:
config::tests::test_project

Indrazar:
Removed the .expect() call by directly appending `"src".into()` to the directories vector, but only when the relative directory is the root of the project.

akesson:
Thanks!

PR #45, opened 2023-01-11
copy server executable to new location before running it

Indrazar:
- solves issues on some platforms where running binaries cannot be edited
- since the binary is copied we need to copy possible debugging files to match


akesson:
Looking good. Thanks!

PR #50, opened 2023-01-14
Fix precedence between config and cli for {bin,lib}-features

nim65s:
Hi,

for `lib-features` and `bin-features`, README says the config  `Can be over-ridden with the command line parameter`.

But running `cargo leptos build --lib-features hydrate,banana --bin-features ssr,apple` on the `start` project compiles fine, while I would expect it to fail because banana and apple are not registered features.

Here is a quick fix, which successfully fail for non-existent features:
```
╰─>$ cargo leptos build --lib-features hydrate,banana --bin-features ssr,apple
error: none of the selected packages contains these features: banana
error: none of the selected packages contains these features: apple
```

### DIFF
--- a/src/config/bin_package.rs
+++ b/src/config/bin_package.rs
@@ -30,10 +30,10 @@ impl BinPackage {
         project: &ProjectDefinition,
         config: &ProjectConfig,
     ) -> Result<Self> {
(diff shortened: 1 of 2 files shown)


akesson:
Yes, that was a bug. Thanks for the contribution!

Issue #53, opened 2023-01-19
`cargo leptos watch` crashes on hot module reloading


SorenHolstHansen:
Tried the `cargo-leptos` setup today, and really liked it, but the hot-module-reloading crashed the page for me. See the video for what happened.

https://user-images.githubusercontent.com/54755687/213550028-042acb7c-3e21-417c-8439-ae53efe47b81.mp4

If it helps, here is my environment
```
System:
    OS: macOS 12.0.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 521.48 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Browsers:
    Firefox: 107.0.1
```

PR #55, opened 2023-01-22
Fix crash when reloading on macos due to code signing issue (#49)

### LINKED ISSUE #49: Watch fails to reload server
Sometimes cargo leptos watch does not start the server. There are logs that seem to indicate that it was attempted.

From discord [thread](https://discord.com/channels/1031524867910148188/1040262908074000425/1063568134746951690)

Issue tracked down to commit https://github.com/leptos-rs/cargo-leptos/commit/94ba14b10a369f3944959c8a3e59ef3dbe3a835e

Trying to summarise: 

- doesn't seem to be a server binary permissions issue.
- unlikely to be a firewall issue
- doesn't output println's from the server, indicating that the binary is never run.
- only happens on Mac
- possibly an ADDRINUSE issue, see [this](https://github.com/watchexec/cargo-watch#reloading-servers-seamlessly)

basilkohler:
Fixes #49 

I certainly is a code signing issue.
I do not completely understand why initially copying the file is ok, but overwriting is not.

I found that on the initial start of `watch` (so after the <bin>_leptos file is copied the first time) if I execute the following command every time before doing any file changes, it would fix the issue:
```
codesign --remove-signature target/server/debug/example_leptos
```
Simply deleting the file before copying also does the trick and I would say is good enough.

### DIFF
--- a/src/service/serve.rs
+++ b/src/service/serve.rs
@@ -77,6 +77,15 @@ impl ServerProcess {
                 GRAY.paint(bin.as_str()),
                 GRAY.paint(new_bin_path.as_str())
             );
+
+            // deleting the old copied binary before overwriting to avoid signing issues on macos
+            if fs::remove_file(&new_bin_path).await.is_ok() {
+                log::debug!(
+                    "Deleted server binary {}",
+                    GRAY.paint(new_bin_path.as_str())
+                );
+            }
+
             fs::copy(bin, &new_bin_path).await?;
             // also copy the .pdb file if it exists to allow debugging to attach
             match determine_pdb_filename(bin) {


Indrazar:
Fix has no regression on Windows 11 22H2 22621.1105

basilkohler:
Should also fix #53

akesson:
Thank you so much for this.

Basically, since windows don't support overwriting a file that is open, which occurs when building, @Indrazar changed it so that the server is running a copy instead. But then it turns out that MacOS 12 get's cranky when overwriting the copy.

Instead of solving that by deleting and then copying, I prefer to only do this copying when on windows. I implemented that on the branch `only-windows-copying` branch and it would be really nice if you could test it to make sure that it doesn't break anything (don't see a reason why but you never know).

basilkohler:
@akesson That is definitely the better approach. Branch works for me (I am still on macOS 12.5).

Indrazar:
`only-windows-copying` branch is working on Windows 11 22H2 22621.1105

Issue #62, opened 2023-01-26
Changes by Tailwind not being actioned.


phillipbaird:
Hi there,

Thank you for creating this cargo plug.  I suspect there is an issue with it working with Tailwinds _watch_ mode.

# Background

TailwindCSS is a utility-first css framework.  It reads the class names referenced in your source files and generates a css file for your site.  Essentially its just a css file generator.  You can run it in _watch_ mode and each time you save a change to your source it updates your css file to match.

# What I'm doing...

I've created a simple starter template for Leptos + Axum + Tailwind that can be found at [https://github.com/phillipbaird/leptos-axum-start](https://github.com/phillipbaird/leptos-axum-start) 

1. Tailwind is started with `tailwindcss -w -i ./tailwind.css -o ./css/main.css`
2. cargo-leptos is started with `cargo leptos watch`
3. Open a browser to localhost:3000.
4. Edit file `app/src/lib.rs` to change a color, e.g. `bg-green-300` is changed to `bg-blue-300`.

# What I expected...

- cargo-leptos observes the change in `app/src/lib.rs` and rebuilds the front end and server.
- Tailwind observes the change and adds `bg-blue-300` to `css/main.css`.
- cargo-leptos observes the change in `css/main.css` and redeploys the file to `target/site/pkg`.
- cargo-leptos notifies the browser of the changes so it can auto reload the changes.

# What actually happened...

It appears as though cargo-leptos is observing the changes but is not actioning them.  Neither the Rust code was rebuilt nor was the css redeployed.

Using `cargo leptos watch -vv` produced the following log when the change was saved...

```bash
listening on 127.0.0.1:3000
      Reload websocket connected
      Notify not handled NoticeWrite("/home/phillipb/Repositories/leptos-experiments/leptos-axum-start/app/src/lib.rs")
      Notify not handled NoticeWrite("/home/phillipb/Repositories/leptos-experiments/leptos-axum-start/css/main.css")
      Notify handle Some("app/src/lib.rs")
      Notify lib source change write "app/src/lib.rs"
      Notify bin source change write "app/src/lib.rs"
   Interrupt send done
          FS create_dir_all "target/site/pkg"
      Assets finished (no changes)
       Style no build needed ChangeSet([LibSource, BinSource])
      Notify handle Some("css/main.css")
      Notify style change write "css/main.css"
   Interrupt send done
       Cargo process interrupted
       Cargo process interrupted
       Build step done with no changes
[...]

PR #65, opened 2023-01-27
Handle interupts to build processes.

phillipbaird:
Fixes https://github.com/leptos-rs/cargo-leptos/issues/62.

When multiple file system updates occur close together in time, the first change (.rs file) may start a build process and the second (Tailwind css) may interrupt the build. The run_loop function was not checking if the
builds were interrupted and would clear the list of file changes. With the notification of the first file change now gone the rebuild process would not complete correctly. This change prevents the change list from being cleared if the builds have been interrupted by additional file system updates.

### DIFF
--- a/src/command/watch.rs
+++ b/src/command/watch.rs
@@ -5,7 +5,7 @@ use crate::{
     config::Project,
     ext::anyhow::Context,
     service,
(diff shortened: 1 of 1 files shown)


akesson:
Thanks

Issue #66, opened 2023-01-30
cargo leptos test - not running integration tests.


phillipbaird:
# Problem Description

Given, a [workspace project structure](https://github.com/phillipbaird/leptos-axum-start) where the `server` contains integration tests in a `tests` directory...

```bash
/app
/front
/server
    /src
    /test
        integration.rs
```

when, you run `cargo leptos test`
then, the integration tests are not run.

# Cause

The [`build_cargo_server_cmd()`](https://github.com/leptos-rs/cargo-leptos/blob/main/src/compile/server.rs#L61-L66) function in `compiler/server.rs` always includes the `--bin` option.
For tests this restricts us to unit tests in the server binary.

# Possible Solution

One possible solution would be to change `build_cargo_server_cmd()` so that `--bin` is not included when running tests.

Thoughts?

akesson:
The integration test folder should be named `tests`, not `test`.
https://doc.rust-lang.org/book/ch11-03-test-organization.html#integration-tests

But I assume that that's only a typo in the description. 

I haven't had much feedback regarding testing. 

There are two possible ways to do tests. The standard rust one (i.e. no target flag set) and the [wasm-bindgen one](https://rustwasm.github.io/wasm-bindgen/wasm-bindgen-test/index.html) where the `--target wasm32-unknown-unknown` is set. The first one is used by cargo-leptos. The second one runs the tests in a browser environment which is very slow and it is experimental. 

Which one are you using?

Assuming the standard rust one.

The `cargo test --help` says:
```
Options:
      --lib                                 Test only this package's library unit tests
      --bin [<NAME>]            Test only the specified binary
     --bins                               Test all binaries
```

Are you saying that when specifying the --bin it doesn't work?

phillipbaird:
> But I assume that that's only a typo in the description.

Your assumption was correct.

> Which one are you using?

It is the standard Rust integration tests for the server that are the issue.

My [starter project](https://github.com/phillipbaird/leptos-axum-start) demonstrates the problem.  If you were to run `cargo leptos test` you will observe that no tests are run.  The one and only integration test in [`tests/integration.rs`](https://github.com/phillipbaird/leptos-axum-start/blob/main/server/tests/integration.rs) is missed.

The reasons the test is missed is because;
- each .rs file in the root of the `tests` directory is treated as as separate crate and compiled into it's own executable and...
[...]

PR #70, opened 2023-02-06
Fix server testing.

phillipbaird:
Changes arguments passed to `cargo test` so that all server test targets are executed.

Resolves https://github.com/leptos-rs/cargo-leptos/issues/66

@akesson This PR depends on an upstream fix in the leptos crate. This is required to resolve an issue with doctests and server functions. Suggest this PR is not released until leptos 0.1.4 is available.


akesson:
Thanks. I'll merge it once the next Leptos release is out.

akesson:
Thanks for the contribution!

PR #71, opened 2023-02-07
Add feature to prevent downloading dependencies

benwis:
To better support Nix, it's good to not download anything during initialization or setup, and let Nix handle any additional deps. This PR throws an error if cargo leptos is built with the no_downloads feature and it needs something.


akesson:
Thanks!

Issue #54, opened 2023-01-20
Hangs on ctrl-c when building


akesson:

phillipbaird:
I think I have a fix for this issue.  You can review this in my fork here: https://github.com/phillipbaird/cargo-leptos/commit/dedddb5549346e190366ddd598e396e0cccebf51

The `build_proj` function did not consider the Outcome of each compile stage. This function has been altered so the build is short circuited if a compile is stopped.

Likewise the `build_all` function would build a list of projects without checking for shutdowns. This function has been changed to check after a call to `build_proj` to also short circuit if a shutdown is requested.

build_proj is called from end2end and at the start of a watch.  Both these modules have been changed so that after `build_proj` has been called they check for shutdown requests and exit if required.

@akesson When you have a moment could you please review and let me know if you'd like a pull request for this?

Issue #72, opened 2023-02-10
cargo leptos watch stop compiling when there is an error in the compiled rust code


danylaporte:
Step to reproduce:
1) start a new project using `cargo leptos --git ...`
2) run the project using `cargo leptos watch`
3) once the project is running, cause a compile error in any *.rs files
4) fix the error (the project does not recompile)

I can reproduce with the --locked version and the bleeding edge. 
I am on Windows 11 with the latest version of rust.

The code seems to stop responding at the following line in src/command/watch.rs

[int.recv().await.dot()?;](https://github.com/leptos-rs/cargo-leptos/blob/25b8a7e8da5043ce7e0d89971ab7128a023baa4c/src/command/watch.rs#L34)

phillipbaird:
# Cause and Effect
1. [wait_interruptible()](https://github.com/leptos-rs/cargo-leptos/blob/main/src/ext/sync.rs#L6-L30) in sync.rs returns a boolean value.  Therefore any callers of this function cannot identify the difference between a failed compile and an interrupted compile as they both are represented by `false`.
2. Both failed and interrupted compiles are mapped to an Outcome::Stopped. See [server.rs](https://github.com/leptos-rs/cargo-leptos/blob/main/src/compile/server.rs#L27-L45) & [front.rs](https://github.com/leptos-rs/cargo-leptos/blob/main/src/compile/front.rs#L34-L36).
3. The run_loop interprets `Outcome::Stopped` as an interrupted compile not a failed compile. It therefore does not clear the change list and so future changes to the same files does not appear to be a change.  And I think this is why the recompile does not happen.

# Possible Solutions

- Introduce a new enum for `wait_interruptible()` to return with variants Success, Fail, Stopped.
Change front.rs and server.rs to map Fail and Stopped to Outcome::Success(Product::None) and Outcome::Stopped respectively.

- Alternatively we could look to reuse the current Outcome enum and maybe add a Failed variant, and return an outcome from `wait_interruptible()`.
For successful compile we'd return Outcome::Success(Product::None)
For failed compile we'd return Outcome::Failed
For interrupted compile we'd return Outcome::Stopped.
This however would then require the caller of `wait_interruptible()` (e.g. server.rs) to then check if the binary has changed and then map Outcome::Success(Product::None) to Outcome::Success(Product::Server).

Happy to create a PR if someone could kindly indicate which solution is preferred.

akesson:
Thanks @danylaporte for reporting this.

[...]

PR #73, opened 2023-02-11
Fix cancelling of builds.

phillipbaird:
This commit is proposed as a solution to https://github.com/leptos-rs/cargo-leptos/issues/54

The `build_proj` function did not consider the Outcome of each compile stage. This function has been altered so the build is short circuited if a compile is stopped.

Likewise the `build_all` function would build a list of projects without checking for shutdowns. This function has been changed to check after a call to `build_proj` to also short circuit if a shutdown is requested.

build_proj is called from end2end and at the start of a watch.  Both these modules have been changed so that after `build_proj` has been called they check for shutdown requests and exit if required.

### DIFF
--- a/src/command/build.rs
+++ b/src/command/build.rs
@@ -8,11 +8,15 @@ use crate::{
         anyhow::{Context, Result},
         fs,
     },
+    signal::{Interrupt, Outcome},
 };
 
 pub async fn build_all(conf: &Config) -> Result<()> {
     for proj in &conf.projects {
         build_proj(proj).await?;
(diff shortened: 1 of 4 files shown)


phillipbaird:
**WARNING**: Just realized that this could exacerbate issue https://github.com/leptos-rs/cargo-leptos/issues/72.
Suggest holding off merging this till a PR is available for that issue.

akesson:
Sorry, saw your comment in #54 with the code in the commit sha and went ahead and adapted that code to the current main (which was updated for #72) and merged it before remembering this PR. 

In any case, thanks!

PR #80, opened 2023-02-23
Make cache dir optional for users building with no_downloads feature

benwis:
I've made some changes to the Exe file to make cache_folder optional, which solves some other problems for NIx users. Everything still seems to work as expected, and all the tests pass.

Love to get another release for cargo-leptos Nix users if everything checks out

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,8 @@ playwright/.cache/
 **/.env
 
 .DS_Store
+
+.direnv
+.envrc
+flake.nix
+flake.lock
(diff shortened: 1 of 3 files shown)


akesson:
Hey @benwis! It seems like the Exe stuff is getting a bit "un-clean" and needs refactoring. This is mainly because of previous changes and contributions. What you are proposing makes sense but I think at this point what would be best is a refactoring so that the cache dir is not needed when creating the enum. 

If you are ok with it, I can go ahead and do that today. I'm just sorry for realizing this now after you already have spent time on this PR.

benwis:
Makes sense to me @akesson! For Nix to be happy, all we need is to not download things and not to create a cache dir. As long as it does that I am happy!

akesson:
Closing this as it was implemented in #81

PR #86, opened 2023-02-26
Fix importing js modules using wasm-bindgen

### LINKED ISSUE #85: External JS modules not being imported properly
Currently external modules with more than one function declared are converted into snippets.

If I were to add the following

```
#[wasm_bindgen(module = "/truffle-wrapper.js")]
extern "C" {
    #[wasm_bindgen(js_name = "setCallback")]
    pub fn set_callback(c: &Closure<dyn Fn(String)>);
    pub fn minimize();
}
```
an import statement is generated within `wasm_bindgen` considering it as an extra module and the following import statement is generated: `import * as __wbg_star0 from './snippets/showdown-hunter-37b34b7f5200334f/truffle-wrapper.js';`

If the module only lists one function like so 
```
#[wasm_bindgen(module = "/truffle-wrapper.js")]
extern "C" {
    #[wasm_bindgen(js_name = "setCallback")]
    pub fn set_callback(c: &Closure<dyn Fn(String)>);
}
```
this import statement is not generated.

In both situations, the module information can be found in the `local_modules` from bindgen as it is retrieved [here](https://github.com/leptos-rs/cargo-leptos/blob/d68c5770968232cae4aa007a36859896267a4697/src/compile/front.rs#L120).

I am not sure but this is possibly an issue on `wasm_bindgen`'s end related to targeting web but I am unsure at this point in time.

SteveCookTU:
This fixes #85

### DIFF
--- a/src/compile/front.rs
+++ b/src/compile/front.rs
@@ -103,7 +103,6 @@ async fn bindgen(proj: &Project) -> Result<Outcome> {
         .input_path(&wasm_file.source)
         .web(true)
         .dot()?
-        .omit_imports(true)
         .generate_output()
         .dot()?;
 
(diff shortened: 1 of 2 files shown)


akesson:
I'll close this PR because I create #92 with more features like:
- detecting if bindgen-produced JS files didn't change in order to reload only when necessary
- watching JS files in order to trigger rebuilds when changed.

I copy-pasted most of the code you did here, so thanks for that.

PR #87, opened 2023-03-02
use cargo-dist to release binaries

ChristopherBiscardi:
This PR uses cargo-dist to generate binaries in a github release **when a tag is pushed to github**.

These binaries are compatible with [cargo-binstall](https://github.com/cargo-bins/cargo-binstall) as well.

The action is generated from `cargo dist init --ci=github` and unmodified from that output. Future updates should be as easy as running this init command again to pull in cargo-dist updates.

cargo-dist recently got a bunch of documentation in the form of a book if you want to know more about what it's doing: https://opensource.axo.dev/cargo-dist/book/ and I've produced a video showing what the results should look like using one of my own projects as an example. https://youtu.be/Cl2qgOwswH4

closes #84

PR #88, opened 2023-03-02
Add binary artifacts with cargo-dist

### LINKED ISSUE #84: [Feature Request] Provide binary artifact
Hi.

Do you plant to provide the binary artifact in the release page?

I am happy to help! 

Thank you for maintaining `cargo-leptos` :heart:

benwis:
Add support of binary uploads on release with cargo-dist. Unfortunately I do not have the ability to release things or run Github Actions, so this will probably work, but I don't know

Fixes #84

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake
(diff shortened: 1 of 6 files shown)


benwis:
Now with prerelease support for arm Macs!

benwis:
Only just now realizing that this is almost the same as #87, with the only difference being that this one also builds binaries for Arm Macs

Issue #90, opened 2023-03-04
end-to-end could report back failure of the npx command


eiswind:
When we run end-to-end (I'm using npx cypress run) and a test fails, the cargo leptos command could report back the failure from the spawned process to make a possible ci pipeline fail.
I tried to extract the ExitStatus in try_run but so far this gets a mess. I guess I'm not form enough with rust to create  an elegant solution.

What I tried is like

```rust
let mut exit_status : Option<ExitStatus> =  None;
tokio::select! {
      _ = int.recv() => {},
     result = process.wait() => {
        if let Ok(status) = result {
            exit_status = Some(status);
        }
     }
}
```
Is there a better way to get hold of the exit status?

eiswind:
After reading the docs with a little more care I worked out

```rust
let result : Option<ExitStatus> = tokio::select! {
      _ = int.recv() => None,
     result = process.wait() => Some(result?)
};
if let Some(exit_status) = result {
    if let Some(exit_code) = exit_status.code() {
        if exit_code != 0 {
            return Err(anyhow!("Command terminated with exit code {}", exit_code))
        }
    }
}
```
I'll check this out in my ci pipeline as soon as I can find time.

akesson:
Closed by #93

PR #93, opened 2023-03-07
respect return code of external test command

eiswind:
https://github.com/leptos-rs/cargo-leptos/issues/90

This pr raises an error when the external test command returns anything other than 0. 

That allows for e.g. a ci pipeline to fail when there have been tests failing.

I am happily awaiting a review !

### DIFF
--- a/src/command/end2end.rs
+++ b/src/command/end2end.rs
@@ -1,6 +1,6 @@
-use std::process::ExitStatus;
 use std::sync::Arc;
 
+use anyhow::bail;
 use camino::Utf8Path;
 use tokio::process::Command;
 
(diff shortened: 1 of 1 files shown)


eiswind:
Maybe we should have a test case?

src/command/end2end.rs
```diff
@@ -49,9 +50,18 @@ async fn try_run(cmd: &str, dir: &Utf8Path) -> Result<()> {
...
+     result = process.wait() => Some(result?)
+    };
+    if let Some(exit_status) = result {
+        if let Some(exit_code) = exit_status.code() {
+            if exit_code != 0 {
+                return Err(anyhow!("Command terminated with exit code {}", exit_code))
+            }
+        }
     }
     Ok(())
```
akesson, line 65: Your code is perfectly fine, but this can also be written as:
```rust
   tokio::select! {
          _ = int.recv() => Ok(()),
          result = process.wait() => {
            let status = result?;
            if !status.success() {
                bail!("Command terminated with exit code {}", status)
            }
            Ok(())
        }
    }
```
I'm fine with both alternatives, so you chose :)

eiswind:
I like your approach. Did run and fail on my ci :)

akesson:
It looks like it passes on the ci here.

eiswind:
Hehe I meant something different I guess. I used the new code to run a playwright test that fails and that did make my (hello-world) pipeline fail, which it did not do before the change.

akesson:
Ahhhh now I understand 🤣 

Ok, looks good. Thanks for the contribution

PR #102, opened 2023-03-14
Always use `/` as separator in css links

### LINKED ISSUE #101: Hot-reloading CSS fails on windows: path sent to the browser via WS contains `\\` separators
The problem is in https://github.com/leptos-rs/cargo-leptos/blob/87ffe4ddab33a498b8b4e53846e6b090fd675628/src/service/reload.rs#L29

`Utf8PathBuf` will use OS specific separators when converting to string.

A solution could look something like this (at least until iter_intersperse is stabilised).
```
// Always use '/' as path separator.
*css_link = file
  .site
  .components()
  .map(|c| c.as_str())
  .collect::<Vec<_>>()
  .join("/");
```

mtsr:
Fix #101

Once iter_intersperse is stabilised, that might be a nicer way to write it.

### DIFF
--- a/src/service/reload.rs
+++ b/src/service/reload.rs
@@ -9,6 +9,7 @@ use axum::{
     routing::get,
     Router,
 };
+use camino::Utf8Component;
 use serde::Serialize;
 use std::sync::Arc;
 use std::{fmt::Display, net::SocketAddr};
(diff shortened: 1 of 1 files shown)


akesson:
Thanks

PR #105, opened 2023-03-29
Readme generated from lib.rs

akesson:


akesson:
The test that produces the readme uses the current working directory, but since it is changed by the test `workspace_build` it fails when the latter runs first.

PR #106, opened 2023-03-29
use relative path to config file for tailwind

Emilgardis:


Emilgardis:
There's one issue with this, the paths in `tailwind.config.js` needs to be adjusted aswell, it's not relative to the config file.

akesson:
thanks

PR #113, opened 2023-04-09
Exclude non-existent paths instead of watching them, fix #112

### LINKED ISSUE #112: superfluously watched src folder in workspaces
Hi,

On the workspace example, I get the following error:
```
┬─ cargo-leptos/examples/workspace on  main via  v1.70.0-nightly at 23:28:23 
╰─ ❯ cargo leptos watch --project project1
    Finished dev [unoptimized + debuginfo] target(s) in 0.10s
       Cargo finished cargo build --package=front-package --lib --target-dir=target/front --target=wasm32-unknown-unknown --no-default-features
    Finished dev [unoptimized + debuginfo] target(s) in 0.13s
       Cargo finished cargo build --package=server-package --bin=server-package --target-dir=target/server --no-default-features
      Notify watching folders project1/css, src, project1/server/src, project1/assets, project1/front/src, project1/app/src
      Notify could not watch "src" due to Io(Os { code: 2, kind: NotFound, message: "No such file or directory" })
```

This doesn't prevent anything to work, but it's a bit misleading.

It looks like this one comes from the `js_dir`, but maybe we can just exclude non-existent paths.

nim65s:


akesson:
Thanks

PR #115, opened 2023-04-13
fix: cause error when a build or test fails

### LINKED ISSUE #114: cargo leptos build returns success error code when wasm32-unknown-unknown target not installed
As above.

It caused me some issues in CI because I forgot to install the target and the build command was silently failing.

markcatley:
fixes #114

### DIFF
--- a/src/command/build.rs
+++ b/src/command/build.rs
@@ -5,16 +5,25 @@ use crate::{
     compile::ChangeSet,
     config::{Config, Project},
     ext::{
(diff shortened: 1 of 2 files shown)


markcatley:
Probably need to do end2end too but I'm not using this on CI at this stage so not 100% sure I'll get it right.

Also, I have retained the current behaviour where `test` and `build` continue after a failure and then exits but it may be best to exit early.

akesson:
Thanks!

Issue #108, opened 2023-04-02
Hot-reloading blocking CSS reload


gbj:
It seems that something in my hot-reloading PR is blocking the style change notification at some point.

I'm assigning this to myself but it will probably a couple weeks before I'm able to take a look so feel free if you have an idea.

## Steps to Reproduce
1. Install latest `cargo-leptos` from git
2. Open Tailwind example in the Leptos repo
3. `cargo leptos watch -v --hot-reload`
4. Edit any class

## Behavior
The class is updated in the DOM, but the CSS file is not reloaded, so it's as if the class has actually disappeared.

`cargo-leptos` logs:
```
      Notify lib source change write "src/app.rs"
      Notify bin source change write "src/app.rs"
      Assets finished (no changes)
       Style no build needed ChangeSet([LibSource, BinSource])
    Patching view.
      Reload server port 127.0.0.1:3000 open
      Reload sent "reload all" to browser
    Blocking waiting for file lock on package cache
      Notify style change write "style/output.css"
       Build interrupted. Restarting.
       Watch waiting for changes
      Assets finished (no changes)
    Blocking waiting for file lock on package cache
    Blocking waiting for file lock on package cache
    Blocking waiting for file lock on package cache

[etc.]

      Reload server port 127.0.0.1:3000 open
      Reload sent "reload all" to browser
```

The new CSS doesn't appear until the reload is set. i.e., Tailwind is updating the CSS file but `cargo-leptos` isn't sending a reload single for the CSS file.

d4h0:
@gbj: With `Tailwind example` you mean [this one](https://github.com/leptos-rs/leptos/tree/main/examples/tailwind), right?

Is this example supposed to use the new Tailwind integration of `cargo-leptos`?

If so, this isn't set up properly.

I was trying to figure out what the issue is here, and was wondering why none of my Tailwind classes were generated – even when I removed the `--hot-reload` flag. Somehow only the initial Tailwind classes, which were defined in the source code on GitHub, worked.

When I opened `Cargo.toml`, I noticed that it doesn't contain the required option to activate Tailwind support.

Besides that, the example includes [`style/output.css`](https://github.com/leptos-rs/leptos/blob/main/examples/tailwind/style/output.css), which is a CSS file that Tailwind has generated.

So basically, to fix the issue, you can do the following:

(1) Delete `style/output.css` (this will automatically be generated and included by `cargo leptos`)

(2) Apply the following changes to `Cargo.toml`

```
[...]

PR #116, opened 2023-04-15
Run style rebuilds separately

gbj:
Addresses #108. At present, if any lib/bin files are touched, the build process waits for style, lib, and bin all to be rebuilt before sending a reload signal, even though it *could* send the style reload signal separately, as this is usually a faster process (Tailwind vs. rustc). 

> Note: This isn't as relevant with plain CSS or with SASS, because in those cases only CSS files are touched and not Rust, so lib/bin aren't recompiled. 

I made this small change to spawn the style rebuild separately rather than joining it with the Rust rebuild.

~However, it only works once, i.e., on first file save. I assume this is because the compilation process that's running in this loop isn't interrupted and stopped if I make another file edit, but I'm not sure. I'm pretty weak on Tokio/concurrent stuff so this may not be the right way to do this.~

Edit: latest commit should actually fix the issue in which it would only recompile styles once, by clearing out the source changes.

### DIFF
--- a/src/command/watch.rs
+++ b/src/command/watch.rs
@@ -65,9 +65,10 @@ pub async fn run_loop(proj: &Arc<Project>) -> Result<()> {
             let proj = Arc::clone(&proj);
             async move {
                 let style = compile::style(&proj, &changes).await;
-                if let Ok(Outcome::Success(Product::Style(_))) = style {
+                if let Ok(Ok(Outcome::Success(Product::Style(_)))) = style.await {
                     ReloadSignal::send_style();
-                    log::info!("Watch updated style")
+                    log::info!("Watch updated style");
+                    Interrupt::clear_source_changes().await;
                 }
             }
         });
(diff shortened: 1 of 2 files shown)


gbj:
I should add: I thought about actually spawning a whole other process with a dedicated watcher but it seemed unnecessary -- however maybe it would've fixed the "only works once" element of this.

akesson:
@gbj sorry, this has been pending for way too long time now. It has some more in-depth changes that would take some time for me to review. However, if you say that it is ok to merge then I'll merge it (with the consequence that if there are regressions, then I trust you to handle them).

gbj:
No worries at all. I think most of that is just noise from accidentally running `cargo fmt`, sorry about that. I think it should be pretty safe to merge, all this really does is just spawn a separate Tokio task for building styles so that updating the stylesheet if you're using Tailwind doesn't block on rebuilding the client and server. The effect of this is that the built-in Tailwind should work with hot-reloading.

akesson:
Ok, I'll go ahead then.

PR #119, opened 2023-05-04
Change repo links from `akesson` to `leptos-rs`.

parker-codes:

PR #120, opened 2023-05-04
[docs] Change `akesson` to `leptos-rs` in Github URLs

dglsparsons:
Also add missing `--locked` to custom install command in README.

(see [discord](https://discord.com/channels/1031524867910148188/1040262908074000425/1103770256738107443) for why). 

I think this is a dupe, but more thorough implementation of #119.


akesson:
Thanks

PR #133, opened 2023-06-07
Support lightningcss 1.0.0-alpha.42

niclas-ahden:
`lightningcss` changed their Rust API as described here: https://github.com/parcel-bundler/lightningcss/releases/tag/v1.21.0

Release 1.21.0 is crate version 1.0.0-alpha.42. Here are the new docs:

- https://docs.rs/lightningcss/1.0.0-alpha.42/lightningcss/printer/struct.PrinterOptions.html
- https://docs.rs/lightningcss/1.0.0-alpha.42/lightningcss/targets/struct.Targets.html

Thanks for building and maintaining this!

PS. I'm not well-versed in Rust and cargo yet, but it might be advisable to tighten the bounds on this dependency if possible. The bound `1.0.0-alpha.40` seems to have allowed version `1.0.0-alpha.42`, which allowed this breaking API-change to come through.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -16,6 +16,18 @@ checksum = "fcb51a0695d8f838b1ee009b3fbf66bda078cd64590202a864a8f3e8c4315c47"
 dependencies = [
  "getrandom",
  "once_cell",
+ "version_check",
+]
+
+[[package]]
+name = "ahash"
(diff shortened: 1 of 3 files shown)


akesson:
Thanks for the help @gbj

PR #140, opened 2023-06-15
Chore: Clippy fixes

martinfrances107:
I am think about starting an issue here..
While I read the code, and get my bearings

it is my habit to run

Cargo clippy
Cargo outdated
Cargo machete

This maybe the first of a few PRs..

"I Ran cargo clippy --fix "

and then reviewed all the changes..

It took me a little time to justify this change
but I see no change in functionality.

```
-    metadata.as_object().map(|o| o.get("leptos")).flatten()
+    metadata.as_object().and_then(|o| o.get("leptos"))
```

Also few .clone() call have been removed which is always nice to see.

### DIFF
--- a/src/compile/style.rs
+++ b/src/compile/style.rs
@@ -31,7 +31,7 @@ pub async fn style(
             log::debug!("Style no build needed {changes:?}");
             return Ok(Outcome::Success(Product::None));
         }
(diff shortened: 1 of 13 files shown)


akesson:
Thanks

PR #141, opened 2023-06-15
Ran cargo update

martinfrances107:


akesson:
Not really necessary, but thanks anyway

Issue #144, opened 2023-06-19
SCSS/SASS compilation fails for Linux MUSL environments


filipdutescu:
# Description

Currently, `cargo-leptos` is unable to compile `.sass`/`.scss` files on non-GNU Linux environments, such as MUSL. For example, to compile a Leptos app for Alpine Linux you'd need to manually get all dependencies and/or install `cargo-leptos` with the `no_downloads` feature. This is less than ideal, as the project could very easily support proper download of the third party dependencies.

# Expected behaviour

`cargo-leptos` downloads the appropriate `dart-sass` (such as [dart-musl/dart-sass](https://github.com/dart-musl/dart-sass)), even on MUSL environments.

# Actual behaviour

`cargo-leptos` downloads the GNU/gclib `dart-sass` which will fail to run/link properly.

# Steps to reproduce

1. Create a basic Leptos app (or use one of the templates) with a `.sass` stylesheet
2. Create a Containerfile/Dockerfile which uses the `rust:alpine` image (it most likely will need some extra dependencies to be installed, such as `ca-certificates npm pkgconfig openssl-dev openssl-libs-static libc-de`)
3. Build the image with Podman/Docker
4. Observe error:
```bash
Command installing wasm-opt-version_112 ...
     Command wasm-opt-version_112 installed.
     Command installing sass-1.58.3 ...
     Command sass-1.58.3 installed.
/root/.cache/cargo-leptos/sass-1.58.3/sass-1.58.3/dart-sass/sass: exec: line 20: /root/.cache/cargo-leptos/sass-1.58.3/sass-1.58.3/dart-sass/src/dart: not found

    Tailwind failed with:
```

PR #145, opened 2023-06-19
feat(dart-musl): download proper dart-sass for musl

filipdutescu:
`dart-sass` is incompatible with MUSL-based Linux, only supporting the standard GNU/glibc version. However there are other third party alternatives which can be used, such as [`dart-musl`][dart-musl].

To accommodate this, if the MUSL `target_env` is found, the SASS download URL is changed to that of the appropriate [`dart-musl`][dart-musl] binary, based on the architecture for which the app is compiled.

[dart-musl]: https://github.com/dart-musl/dart-sass

Closes: #144

### DIFF
--- a/src/ext/exe.rs
+++ b/src/ext/exe.rs
@@ -10,7 +10,7 @@ use std::{
 };
 use zip::ZipArchive;
 
-use super::util::os_arch;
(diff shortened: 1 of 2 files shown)


akesson:
Thanks!

PR #151, opened 2023-07-07
Cargo wrappers

### LINKED ISSUE #149: Support for cargo wrappers (like cross)
It would be useful to be able to use cross to compile the server for different targets. I think supporting this could be implemented by making the name of the "cargo" executable used to run the compilation commands configurable, thereby supporting not only cross but anything that wraps cargo.

Raniz85:
Suppor for alternate commands instead of 'cargo' when building the server as well as configuring the target directory for the server.

Closes #149


akesson:
Changes look good. I didn't test it, but I'll trust you!

PR #153, opened 2023-07-07
Update README.md to list starter templates

gbj:
The README currently contains an old note about asking in the Discord for a template link, from when we didn't really have them ready yet. This replaces it with links to the typical starters.


akesson:
Thanks!

PR #154, opened 2023-07-07
Use LEPTOS_BIN_TARGET_DIR for exe file

Raniz85:


Raniz85:
Didn't catch this usage until I compiled in a clean workspace :/

Raniz85:
It seems the new test is flaky due to parallell running, I'll see if I can come up with something to fix it.

Raniz85:
I'll get back to this on monday - weekend starts now... Feel free to revert #151 if the tests are causing you grief in the meantime.

Raniz85:
@gbj I removed the flaky test because getting it to work would require significant changes to how environment variables are picked up to be able to inject them in the tests without setting the environment for all tests.

Running with one test runner would also be a solution, but I don't like non-parallel tests for several reasons.

akesson:
Yes, there's an important refactoring that I just don't find time to do to remove the use of CWD which screws up testing.

PR #159, opened 2023-07-17
Respect Cargo target directory configuration

### LINKED ISSUE #74: Panic when `CARGO_TARGET_DIR` is not "target"
Panic:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Could not remove base "/tmp/start" from "/tmp/it_doesnt_really_matter"', /home/aru/.cargo/registry/src/github.com-1ecc6299db9ec823/cargo-leptos-0.1.7/src/ext/cargo.rs:73:68
```

Simple reproduction on linux:
```sh
git clone https://github.com/leptos-rs/start /tmp/start
cd !$
CARGO_TARGET_DIR=/tmp/it_doesnt_really_matter cargo leptos build
```

### LINKED ISSUE #127: cargo leptos fails on custom build target-dir
Hey, seems like I'm unable to compile examples having a custom target-dir set up. This is my config in `/home/foo/.cargo/config`:
```
[build]
target-dir = ".cargo/target"
```

and following is the error being thrown on `cargo leptos`:
```sh
~/github/leptos/examples/hackernews_axum ❯ cargo leptos watch
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Could not remove base "/home/foo/github/leptos/examples/hackernews_axum" from "/home/foo/.cargo/target"', /home/foo/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cargo-leptos-0.1.8/src/ext/cargo.rs:73:68
```

### LINKED ISSUE #157: thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Could not remove base
image
image
image

### LINKED ISSUE #186: `cargo leptos watch` fails when CARGO_TARGET_DIR is defined
When CARGO_TARGET_DIR is defined to redirect target files to a single folder (so we don't have target folders all over out harddisks), the command `cargo leptos watch` fails.

You can recreate the problem thus:

```
mkdir ~/.cargo-target
export CARGO_TARGET_DIR=~/.cargo-target
... create a new project via cargo leptos new
cd <new-folder>
cargo leptos watch
```

tyoeer:
Changes:
- Removes hard-coded `target/server` and `target/front`
	- Using the normal target directory allows for sharing build dependencies between projects, which can significantly reduce compile times
	- It is unclear why they're hardcoded separate directories in the first place
	- Removing them was easier than making them use the proper target directory
- Uses proper relative paths for bin and lib packages
	- Fixes a crash/panic when the configured target directory was outside the workspace directory
		- Fixes #74
		- Fixes #127
		- Fixes #157
		- Fixes #186
	- Adds dependency `pathdiff`
		- It can still fail, but insofar I can tell will only do so if the base path is relative and the target path is absolute
	- Absolute paths would be better:
		- No chance of failure
		- No additional dependency
		- But those break snapshot testing
- Defaults `site-root` to the `/site/` subfolder of the cargo target directory
	- That is insofar I can tell how it was intended
	- The current method is a bit clunky, but it's the best I could come up with with serde's default
		- An alternative would be to use an option and set it to `Some` in `parse()`,
		  but that would forces `Some` checks everywhere.
	- Snapshot testing survives this because the test cases always manually specify it.

Didn't manage to add additional regression tests:
- Changing the target directory has to be done through a `.cargo/config.toml` from where cargo is invoked,
  and does not take the `--manifest-path` into account
	- Known problem: https://github.com/rust-lang/cargo/issues/10098
- The tests already specify a `cwd`, though it'sn't used for the `cargo metadata` call,
  and isn't set to the directory of the actual project/workspace being tested. I can't figure out what exactly it's supposed to do,
  and I don't feel confident doing something with it.

Additional notes/things to look at that are outside the scope of this PR:
- Snapshot testing is broken for developers who have a different Cargo target directory
- clippy has complaints about cloning double references in `src/ext/path.rs`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -370,6 +370,7 @@ dependencies = [
  "lightningcss",
  "log",
  "notify",
+ "pathdiff",
  "reqwest",
  "seahash",
  "semver",
@@ -1778,6 +1779,9 @@ name = "pathdiff"
 version = "0.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8835116a5c179084a830efb3adc117ab007512b535bc1a21c991d3b32a6b44dd"
+dependencies = [
+ "camino",
+]
 
 [[package]]
 name = "percent-encoding"
--- a/src/config/bin_package.rs
+++ b/src/config/bin_package.rs
@@ -90,7 +90,7 @@ impl BinPackage {
             };
             let mut file = config.bin_target_dir.as_ref()
                 .map(|dir| dir.into())
-                .unwrap_or_else(|| metadata.rel_target_dir().join("server"));
+                .unwrap_or_else(|| metadata.rel_target_dir());
             if let Some(triple) = &config.bin_target_triple {
                 file = file.join(triple)
             };
--- a/src/config/bin_package.rs
+++ b/src/config/bin_package.rs
@@ -90,6 +90,7 @@ impl BinPackage {
             };
             let mut file = config.bin_target_dir.as_ref()
                 .map(|dir| dir.into())
+                // Can't use absolute path because the path gets stored in snapshot testing, and it differs between developers
                 .unwrap_or_else(|| metadata.rel_target_dir());
             if let Some(triple) = &config.bin_target_triple {
                 file = file.join(triple)
--- a/src/config/lib_package.rs
+++ b/src/config/lib_package.rs
@@ -67,7 +67,6 @@ impl LibPackage {
         let wasm_file = {
             let source = metadata
                 .rel_target_dir()
-                .join("front")
                 .join("wasm32-unknown-unknown")
                 .join(profile.to_string())
                 .join(name.replace('-', "_"))
--- a/src/config/lib_package.rs
+++ b/src/config/lib_package.rs
@@ -66,6 +66,7 @@ impl LibPackage {
 
         let wasm_file = {
             let source = metadata
+                // Can't use absolute path because the path gets stored in snapshot testing, and it differs between developers
                 .rel_target_dir()
                 .join("wasm32-unknown-unknown")
                 .join(profile.to_string())
--- a/src/config/project.rs
+++ b/src/config/project.rs
@@ -180,16 +180,25 @@ pub struct ProjectConfig {
 }
 
 impl ProjectConfig {
-    fn parse(dir: &Utf8Path, metadata: &serde_json::Value) -> Result<Self> {
+    fn parse(dir: &Utf8Path, metadata: &serde_json::Value, cargo_metadata: &Metadata) -> Result<Self> {
         let mut conf: ProjectConfig = serde_json::from_value(metadata.clone())?;
         conf.config_dir = dir.to_path_buf();
         let dotenvs = load_dotenvs(dir)?;
         overlay_env(&mut conf, dotenvs)?;
-        if conf.site_root == "/" || conf.site_root == "." {
+        if conf.site_root == "/" || conf.site_root == "." || conf.site_root == "CARGO_TARGET_DIR" {
             bail!(
                 "site-root cannot be '{}'. All the content is erased when building the site.",
                 conf.site_root
             );
+        }
+        if conf.site_root.starts_with("CARGO_TARGET_DIR") {
+            conf.site_root = {
+                let mut path = cargo_metadata.target_directory.clone();
+                // unwrap() should be safe because we just checked
+                let sub = conf.site_root.unbase("CARGO_TARGET_DIR".into()).unwrap();
+                path.push(sub);
+                path
+            };
         }
         if conf.site_addr.port() == conf.reload_port {
             bail!(
@@ -212,11 +221,12 @@ impl ProjectDefinition {
     fn from_workspace(
         metadata: &serde_json::Value,
         dir: &Utf8Path,
+        cargo_metadata: &Metadata,
     ) -> Result<Vec<(Self, ProjectConfig)>> {
         let mut found = Vec::new();
         if let Some(arr) = metadata.as_array() {
             for section in arr {
-                let conf = ProjectConfig::parse(dir, section)?;
+                let conf = ProjectConfig::parse(dir, section, cargo_metadata)?;
                 let def: Self = serde_json::from_value(section.clone())?;
                 found.push((def, conf))
             }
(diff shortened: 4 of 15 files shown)


tyoeer:
Had to temporarily remove my own `target-dir` configuration to make it match the CI, but the tests look like they should match what the CI is expecting now.

Cargo.lock
aperepel, line 1: Do you mind rebasing to the latest main branch? This should only have the 1 lib update ideally, causes a manual conflict every time. Thanks!

src/config/bin_package.rs
```diff
@@ -90,7 +90,8 @@ impl BinPackage {
             };
             let mut file = config.bin_target_dir.as_ref()
                 .map(|dir| dir.into())
-                .unwrap_or_else(|| metadata.rel_target_dir().join("server"));
+                //Can't use absolute path because the path gets stored in snapshot testing, and it differs between developers
```
aperepel, line 93: Could you please update to follow the rust fmt guidelines? (space after the opening slashes).

src/config/lib_package.rs
```diff
@@ -66,8 +66,8 @@ impl LibPackage {
 
         let wasm_file = {
             let source = metadata
+                //Can't use absolute path because the path gets stored in snapshot testing, and it differs between developers
```
aperepel, line 69: Rust fmt: https://doc.rust-lang.org/book/ch03-04-comments.html#:~:text=In%20Rust%2C%20the%20idiomatic%20comment,%23!%5B

src/config/project.rs
```diff
@@ -180,17 +180,26 @@ pub struct ProjectConfig {
...
 
 impl ProjectConfig {
-    fn parse(dir: &Utf8Path, metadata: &serde_json::Value) -> Result<Self> {
+    fn parse(dir: &Utf8Path, metadata: &serde_json::Value, cargo_metadata: &Metadata) -> Result<Self> {
         let mut conf: ProjectConfig = serde_json::from_value(metadata.clone())?;
         conf.config_dir = dir.to_path_buf();
         let dotenvs = load_dotenvs(dir)?;
         overlay_env(&mut conf, dotenvs)?;
-        if conf.site_root == "/" || conf.site_root == "." {
+        if conf.site_root == "/" || conf.site_root == "." || conf.site_root == "CARGO_TARGET_DIR" {
```
aperepel, line 188: Should this include the synonym `CARGO_BUILD_TARGET_DIR` as well? https://doc.rust-lang.org/cargo/reference/config.html#buildtarget-dir

src/config/project.rs
```diff
@@ -180,17 +180,26 @@ pub struct ProjectConfig {
...
             bail!(
                 "site-root cannot be '{}'. All the content is erased when building the site.",
                 conf.site_root
             );
         }
+        if conf.site_root.starts_with("CARGO_TARGET_DIR") {
```
aperepel, line 194: Add support for the synonym env var https://doc.rust-lang.org/cargo/reference/config.html#buildtarget-dir

src/config/project.rs
```diff
@@ -180,17 +180,26 @@ pub struct ProjectConfig {
...
+            conf.site_root = {
+                let mut path = cargo_metadata.target_directory.clone();
+                //unwrap should be safe because we just checked
+                let sub = conf.site_root.unbase("CARGO_TARGET_DIR".into()).unwrap();
```
aperepel, line 198: Add synonym. Also, extract to a constant? More than a few places the env var name is used by now :)

src/config/project.rs
```diff
@@ -285,7 +296,7 @@ fn default_pkg_dir() -> Utf8PathBuf {
 }
 
 fn default_site_root() -> Utf8PathBuf {
-    Utf8PathBuf::from("target").join("site")
+    Utf8PathBuf::from("CARGO_TARGET_DIR").join("site")
```
aperepel, line 299: Add synonym, too https://doc.rust-lang.org/cargo/reference/config.html#buildtarget-dir

aperepel (changes requested):
Thank you for this PR. I've added some suggestions, the primary one being to add support for the synonym var name as per https://doc.rust-lang.org/cargo/reference/config.html#buildtarget-dir

As a general comment, could you please include a sample dev experience walkthrough of how to leverage these improvements? This way I don't have to guess if I did the right set up to verify changes are doing what they are intended to accomplish.

src/config/project.rs
tyoeer: @aperepel This function only instantiates a _single_ default, adding a synonym/second path variant doesn't make sense here (to me). Did you mean that it should simply be replaced with `"CARGO_BUILD_TARGET_DIR"` (or the appropriate constant)?

tyoeer:
> As a general comment, could you please include a sample dev experience walkthrough of how to leverage these improvements? This way I don't have to guess if I did the right set up to verify changes are doing what they are intended to accomplish.

@aperepel 
The easiest way to test everything would be to configure the target directory in the top-level cargo configuration in `~/.cargo/config.toml`, for example with:
```toml
[build]
target-dir = ".cargo/targetShare"
```
This shares the target directory between projects. I personally do this to speed up compilation and save disk-usage by re-using already compiled dependencies (between projects), other people might have other reasons.

With this configuration, one would expect that:
- No panics happen
  - Previously it would panic for target directories outside of the workspace (see linked issues)
- By default, the server binary gets built in the configured target directory, and can reuse compiled dependencies
  - This is configurable, though previously it would default to `target/server`, irregardless of the configured target directory, effectually overwriting your config for no apparent reason.
- The wasm/client binary gets built in the configured target directory (though in the ` wasm32-unknown-unknown` subfolder), and can reuse compiled dependencies
  - Previously, this was _hard-coded_ to `target/front`, with no way to configure/change it, , effectually overwriting your target directory config for no apparent reason.
- By default, the other site assets are put somewhere in the configured target folder
  - Previously, this would default to `target/site`, irregardless of the configured target directory, effectually overwriting your config for no apparent reason.

(Ideally there'd be a test demonstrating all this, but see my original PR comment for why I didn't manage to make one)

Maneren:
Hello, what's the status here? I've been using a binary compiled from this PR for quite some time now (without any issues) and I'd love to see this merged so I can use other updates as well. Is there anything I can help with? (code, testing, review, ...)

gbj:
Sorry for the delay here — I've been reviewing PRs post-0.5 and am going to start going through these now.

PR #165, opened 2023-07-23
Implement precompression of static assets

sebadob:
This is about adding a build step for `cargo-leptos` which does an optional precompression of the static assets in release build mode.

This provides a huge performance boost both on the server, which does not need to dynamically compress static files for each request, and in transit to the client, if the server does not do compression in some middleware.
Additionally, it allows an easy embedding of precompressed static assets into the final binary itself, which improves the performance even further.

I have a first (still incomplete) implementation and will update the description after some testing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ playwright/.cache/
 **/.env
 
 .DS_Store
+
+.idea/
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,6 +9,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
+name = "adler32"
+version = "1.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "aae1277d39aeec15cb388266ecc24b11c80469deae6067e17a1a7aa9e5c1f234"
+
+[[package]]
 name = "ahash"
 version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -39,6 +45,21 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "43f6cb1bf222025340178f382c426f13757b2960e89779dfcb319c32542a5a41"
 dependencies = [
  "memchr",
+]
+
+[[package]]
+name = "alloc-no-stdlib"
+version = "2.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cc7bb162ec39d46ab1ca8c77bf72e890535becd1751bb45f64c597edb4c8c6b3"
+
+[[package]]
+name = "alloc-stdlib"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "94fb8275041c72129eb51b7d0322c29b8387a0386127718b096429201a5d6ece"
+dependencies = [
+ "alloc-no-stdlib",
 ]
 
 [[package]]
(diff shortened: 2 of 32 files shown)


sebadob:
I did some testing today and used this version instead of the "original" `cargo-leptos` for the development and it works out very nicely so far.

The idea is, that you can specify the `-P` or `--precompress` flag, when you do a `build -r` or `serve -r`. This will precompress all static assets with gzip (for compatibility) and brotli. This greatly reduces the assets sizes of course and since these are static files, there is no need for a framework like axum or actix to do this compression each time over and over again with each request, when the results are static anyway.

What I am doing on my project is that am using `rust_embed` for embed the precompressed assets into the final binary for even better performance. You can either serve the precompressed files with a manual handler, for instance something like this:

```rust
use axum::body::Body;
use axum::extract::State;
use axum::{
    body::{boxed, Full},
    http::{header, Response, StatusCode},
    response,
};
use common::constants::DEV_MODE;
use http::{Request, Uri};
use leptos::*;
use std::borrow::Cow;
use tracing::{error, info};

#[derive(rust_embed::RustEmbed)]
#[folder = "../target/site/"]
struct Assets;

pub async fn file_and_error_handler(
    uri: Uri,
    State(_options): State<LeptosOptions>,
    req: Request<Body>,
) -> response::Response {
    let (_, path) = uri.path().split_at(1); // split off the first `/`
    let mime = mime_guess::from_path(path);

    let accept_encoding = req
        .headers()
        .get("accept-encoding")
        .map(|h| h.to_str().unwrap_or("none"))
        .unwrap_or("none");
    let (path, encoding) = if *DEV_MODE {
        // during DEV, don't care about the precompression -> faster workflow
        (Cow::from(path), "none")
    } else if accept_encoding.contains("br") {
        (Cow::from(format!("{}.br", path)), "br")
    } else if accept_encoding.contains("gzip") {
        (Cow::from(format!("{}.gz", path)), "gzip")
    } else {
        (Cow::from(path), "none")
    };

    match Assets::get(path.as_ref()) {
        Some(content) => {
            let body = boxed(Full::from(content.data));

            match *DEV_MODE {
                true => Response::builder()
                    .header(header::CONTENT_TYPE, mime.first_or_octet_stream().as_ref())
                    .header(header::CONTENT_ENCODING, encoding)
                    .body(body)
                    .unwrap(),
                false => {
                    Response::builder()
                        .header(header::CACHE_CONTROL, "max-age=86400")
                        .header(header::CONTENT_TYPE, mime.first_or_octet_stream().as_ref())
                        .header(header::CONTENT_ENCODING, encoding)
                        .body(body)
                        .unwrap()
                }
            }
        }
        
        None => {
            error!(">> Asset {} not found", path);
            for a in Assets::iter() {
                info!("Available asset: {}", a);
            }
            Response::builder()
                .status(StatusCode::NOT_FOUND)
                .body(boxed(Full::from("not found")))
                .unwrap()
        }
    }
}

```

Or if you do not want to directly embed the assets, with axum you can simply add somehting like this in your router:

```rust
ServeDir::new("../target/site/")
    .precompressed_br()
    .precompressed_gzip(),
```

For actix, there is an open issue to make it work just like with axum.

You can build or serve with this feature easily with `cargo leptos build -r -P`, which will then result in something like this:

Bildschirmfoto vom 2023-07-24 14-40-51

This precompression will be applied for `--release` builds only. It does not serve any advantage in local development, it would just slow down recompliation.

thestarmaker:
Afternoon! This seem slike a very cool feature, what's the status on this?

sebadob:
It's working perfectly fine, ready for review / merge.
I just don't have the rights to merge it into master.

gbj:
Sorry for the long delay! I've been head down on 0.5 in my Leptos time for a while now, but have been reviewing this and the other open PRs and will be releasing a new version soon. Thanks so much for your work.

thestarmaker:
@sebadob I was thinking about you earlier on, this is a wonderful feature (I was probably one of the first to actually plug it into my project before this was even merged). It seems like we have reached the day when `zstd` is supported by 74% of browsers globally: https://caniuse.com/zstd.

Would you be interested in adding `zstd` support into precompression feature?
(If not, I may try myself, but I may butcher it 😅 )

thestarmaker:
actually, I have just checked compression ratios, it seems zstd is only marginally better (compression ratio) than brotly when both are cranked up to max. So maybe not... 😞

PR #166, opened 2023-07-24
Fixed musl binary download for cargo-generate

rdbo:
Added detection for Linux targets that use the musl libc on the download of the `cargo-generate` binary. The current `main` branch does not do any detection for it, only for `SASS`.
With this pull request, I managed to get cargo-leptos and leptos itself fully working on Alpine Linux.

### DIFF
--- a/src/ext/exe.rs
+++ b/src/ext/exe.rs
@@ -452,13 +452,23 @@ impl Command for CommandCargoGenerate {
     fn github_repo(&self) -> &'static str { "cargo-generate" }
 
     fn download_url(&self, target_os: &str, target_arch: &str, version: &str) -> Result<String> {
-        let target = match (target_os, target_arch) {
-            ("macos", "aarch64") => "aarch64-apple-darwin",
-            ("linux", "aarch64") => "aarch64-unknown-linux-gnu",
-            ("macos", "x86_64") => "x86_64-apple-darwin",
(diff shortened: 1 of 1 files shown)


aperepel:
Leaving a note here to review when #168 is merged, as I did extensive refactoring of the commands there.

rdbo:
> Leaving a note here to review when #168 is merged, as I did extensive refactoring of the commands there.

I could make another PR when yours get merged. Since I use a Linux with musl, I could also test mine and your changes right away.

aperepel:
Hey @rdbo , the #168 has been merged. Feel free to update this PR, the changes will go into the https://github.com/leptos-rs/cargo-leptos/blob/3320ec6f029482cf292853edec64fb582341385d/src/ext/exe.rs#L496 function impls.

I don't have a musl-based linux readily accessible at the moment, but looking over the changes, was wondering - should this logic be applied to other commands, too? Would appreciate if you could check if it applies to other command binaries and update. Thanks!

rdbo:
There already is a check for `SASS` for musl, but not for `cargo-generate`. As for the other binaries, they seem to be statically compiled, since they are very big. I will double check though.
Working on the PR.

rdbo:
@aperepel rebased from your changes and added the proper `cargo-generate` binary download for `musl`. Everything is fine, I got the `leptos-rs/start` template to work :+1: 
As for the other binaries, they are statically linked, so they're the same on glibc and musl.

PR #167, opened 2023-07-25
Exit serve command if child process exits

stewartadam:
Currently, `serve` and `watch` share an underlying server which causes `serve` to wait for ctrl+c to be pressed, even if the underlying process has gracefully exited.

This PR makes `serve` behave as a one-shot and exit if its underlying process terminates, while preserving the reload functionality of the `watch` command.

### DIFF
--- a/src/command/serve.rs
+++ b/src/command/serve.rs
@@ -8,7 +8,7 @@ pub async fn serve(proj: &Arc<Project>) -> Result<()> {
     if !super::build::build_proj(proj).await.dot()? {
         return Ok(());
     }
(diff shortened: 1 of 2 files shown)


stewartadam:
Tangentially related; what are your thoughts of aliasing cargo-leptos' `serve` command as `run`? Provided that `serve` was indeed intended to be a one-shot run, it would be intuitive to match the native `cargo` command syntax.

gbj:
Thanks, this looks like a nice small improvement! 

I'm fine with the idea of aliasing `serve` as `run`.

PR #171, opened 2023-07-28
Minor housekeeping. ran cargo update.

martinfrances107:
external libraries 6 removed, 9 added

internal libraries
 leptos-hot-reload updated.

Also in semiverison numbering terms many "patch level" updates.

When it happens, troubleshooting a

> "Which minor level patch change triggered a compilation failure"

issue is a major headache. Periodic updates here are a potential life-saver.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,15 @@
 version = 3
 
 [[package]]
+name = "addr2line"
+version = "0.20.0"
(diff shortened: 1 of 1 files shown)


aperepel:
Thank you for this PR. I re-ran it to pick a few more patch updates since it was submitted, committing next.

PR #179, opened 2023-08-02
Dropped dependency on once_cell

martinfrances107:
In rust version 1.70.x "OnceCell" and it sync equivalent "OnceLock" were stabilized and moved in "std"

https://blog.rust-lang.org/2023/06/01/Rust-1.70.0.html

From a maintenance perspective - One less package to track. From a security perspective - A reduced attack surface.

I have updated the Cargo.tomls "rust-version" to 1.70 to enforce this new reality.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -370,7 +370,6 @@ dependencies = [
  "lightningcss",
  "log",
  "notify",
- "once_cell",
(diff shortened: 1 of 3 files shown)


aperepel:
Resolved via a combined #181 . Thank you!

PR #180, opened 2023-08-02
Rev up the project version to 0.1.12

aperepel:


aperepel:
Need not merge just yet, depending on the release workflow. `cargo release` can rev up the version (or use the one set explicitly). Will need to clarify the team workflow there.

PR #183, opened 2023-08-11
Fix debug profile in readme

s0l0ist:
The readme incorrectly describes how to specify a debug/dev profile.

The correct way is to use the `dev` key word.


gbj:
Thanks. It looks like we renamed it at some point. 
https://github.com/leptos-rs/cargo-leptos/blob/010b65fd091769233fe2bb8ec068b05d6165b8d8/src/config/profile.rs#L4-L8

Good catch and thanks for the PR.

PR #189, opened 2023-08-24
Fix documentation for end-to-end command

CharlesTaylor7:
It seems like the command was renamed at some point from `end2end` to `end-to-end`.


gbj:
Thanks very much, and apologies for the slowness on my end!

PR #197, opened 2023-10-04
Update Rust to make a release.

lephyrius:
Updates the rust version to make it pass github actions.


gbj:
Thanks so much! I ran out of steam to fix this last night.

PR #200, opened 2023-10-06
Fix tailwind with plugins by using a temporary file

### LINKED ISSUE #136: Failing for tailwind plugins
Build (and watch) fails after adding plugins to `tailwind.config.js`

```js
module.exports = {
  content: {
    files: ["*.html", "./src/**/*.rs"],
  },
  plugins: [require("@tailwindcss/typography"), require("daisyui")],
};
``` 
Screenshot 2023-06-11 at 18 24 35

I've installed all required dependencies locally with `npm i -D tailwindcss @tailwindcss/typography daisyui`

crapStone:
closes #136 

Let tailwind output to a temporary file and read it afterwards instead of returning the stdout of the tailwind command.

I didn't know a better way than to put the temp file into a `tmp` directory inside the `target` folder.
`temp_dir` isn't supported by serde and I didn't want to put it into the `TailwindConfig` struct.

The path is present in the output for the output of the tailwind command for better transparency with the users.

### DIFF
--- a/src/compile/tailwind.rs
+++ b/src/compile/tailwind.rs
@@ -34,7 +34,13 @@ pub async fn compile_tailwind(
 
             if done {
                 log::info!("Tailwind finished {}", GRAY.paint(line));
-                Ok(Outcome::Success(output.stdout()))
+                match fs::read_to_string(&tw_conf.tmp_file).await {
+                    Ok(content) => Ok(Outcome::Success(content)),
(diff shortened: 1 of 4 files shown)


gbj:
Thanks very much! I appreciate these recent contributions.

ckissane:
Thank you! Good workaround!

PR #202, opened 2023-10-06
Add 'node_modules/.bin' folder to PATH when package.json is present

crapStone:
`cargo leptos` only searches for globally installed binaries.

This PR adds a check at the startup of `cargo leptos` that detects a `package.json` file and tries to add the `<project_working_dir>/node_modules/.bin/` folder to the `PATH` environment variable.


gbj:
Looks good, thanks!

PR #203, opened 2023-10-07
Fix incremental compilation by using separate target dir for front

crapStone:
`cargo leptos watch` compiles frontend and server code completely new, without using incremental compilation.

This is a bug of `cargo` or `rustc`, because changing the target from `native` to `wasm32` changes the `RUSTFLAGS` which in turn triggers a complete rebuild by cargo (see https://github.com/rust-lang/cargo/issues/8716).

This PR introduces a new option in the project settings (`separate-front-target-dir`) that instructs `cargo leptos` to change the `--target-dir` for the frontend compilation to `<project-taget-dir>/front` so that the `RUSTFLAGS` stay the same between the compilations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,6 +188,11 @@ lib-profile-release = "my-release-profile"
 #
 # Optional. Defaults to "debug".
 lib-profile-dev = "my-debug-profile"
+
+# Fixes cargo bug that prevents incremental compilation (see #203)
+#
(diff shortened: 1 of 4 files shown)


OvermindDL1:
Just as a passing question, shouldn't this just have been a new cargo profile instead of a new cargo target directory?

PR #204, opened 2023-10-12
Add support for cargo-binstall

azzamsa:


gbj:
`cargo binstall cargo-leptos` already works (at least for me). [Per `cargo-binstall` docs](https://github.com/cargo-bins/cargo-binstall/blob/main/SUPPORT.md):

> To get started, check the [default](https://github.com/cargo-bins/cargo-binstall/blob/main/SUPPORT.md#Defaults) first, only add a [package.metadata.binstall] section to your Cargo.toml if the default does not work for you.

Is something not working?

azzamsa:
Oh, It's news to me. I thought there must be some kind of "default" metadata in Cargo.toml

PR #208, opened 2023-10-27
Add bin_cargo_args and lib_cargo_args to cargo-leptos

benwis:
I wanted to pass additional parameters to the cargo processes used by cargo-leptos, primarily --timings, so I added an option to pass additional cargo flags in the CLI and as a setting in Cargo.toml.


gbj:
Thanks, Ben! Looks good to me. I'm not sure what's up with the Windows tests but it's happening on the main branch too.

Issue #129, opened 2023-05-23
Watch command fails due to incompatible wasm bindgen schema


adsick:
I tried updating everything many times but it won't help.

```cli
leptos/examples/counter_isomorphic on main [⇣] is 📦 v0.1.0 via 🦀 v1.71.0-nightly took 2m8s 
❯ cargo leptos watch
    Finished dev [unoptimized + debuginfo] target(s) in 0.10s
       Cargo finished cargo build --package=counter_isomorphic --lib --target-dir=target/front --target=wasm32-unknown-unknown --no-default-features --features=hydrate
Error: at `/home/adsick/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cargo-leptos-0.1.8/src/compile/front.rs:42:30`

Caused by:
    0: at `/home/adsick/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cargo-leptos-0.1.8/src/compile/front.rs:108:10`
    1: 
       
       it looks like the Rust project used to create this wasm file was linked against
       version of wasm-bindgen that uses a different bindgen format than this binary:
       
         rust wasm file schema version: 0.2.86
            this binary schema version: 0.2.84
       
       Currently the bindgen format is unstable enough that these two schema versions
       must exactly match. You can accomplish this by either updating the wasm-bindgen
       dependency or this binary.
       
       You should be able to update the wasm-bindgen dependency with:
       
           cargo update -p wasm-bindgen
       
       or you can update the binary with
       
           cargo install -f wasm-bindgen-cli
       
       if this warning fails to go away though and you're not sure what to do feel free
       to open an issue at https://github.com/rustwasm/wasm-bindgen/issues!
```

adsick:
also a little note: that link at the end has an adjacent exclamation which gets inserted when you <kbd>Ctrl</kbd>+<kbd>click</kbd> on it in VS code leading to a 404 and confusion.

jsprog:
Already faced this problem and got it resolved by:
- `cargo uninstall cargo-leptos`
- `cargo install --git https://github.com/leptos-rs/cargo-leptos --locked cargo-leptos`
- `cargo clean`

adsick:
@jsprog actually only
```cargo install --git https://github.com/leptos-rs/cargo-leptos --locked cargo-leptos```
is enough to make it work, thanks. I believe we need to either mention that in the readme, or always make sure that cargo-leptos is always up to date on crates.io.

[...]

PR #211, opened 2023-11-01
cargo update

nim65s:
Hi,

This is just a `cargo update` for wasm-windgen [v0.2.88](https://github.com/rustwasm/wasm-bindgen/releases/tag/0.2.88), to fix an issue similar to #129

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -305,9 +305,9 @@ dependencies = [
 
 [[package]]
 name = "brotli-decompressor"
-version = "2.5.0"
+version = "2.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "da74e2b81409b1b743f8f0c62cc6254afefb8b8e50bbfe3735550f7aeefa3448"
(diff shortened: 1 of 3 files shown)


JoelBeicher:
Hey @nim65s I think those checks aren't going through due to the example projects.
The need to be updated to use the current leptos version 0.5, I think.

gbj:
I think maybe you need to run `cargo update` in the two projects, so that they use wasm-bindgen 0.2.88 as well. You can see that their lockfiles have them at 0.2.87 still

https://github.com/leptos-rs/cargo-leptos/blob/0fbe4bb7c6773c696866eff5530c6c432bacc3fa/examples/project/Cargo.lock#L2820

nim65s:
Thanks ! Let's try this.

nim65s:
Now we have a test failure on windows related to path separators, but this test was already failing before this PR.

gbj:
Thanks, looks good! Agreed we need to fix the Windows thing but it's not this PR.

PR #212, opened 2023-11-10
Add --bin-skip flag to build command

CorvusPrudens:
## Utility

In certain pipelines, it is convenient to use `cargo-leptos` to build the lib target and associated files only. This can be done currently by re-implementing the functionality `cargo-leptos` provides manually, but that may require parsing the project's `Cargo.toml`, and it certainly requires more knowledge of the build process than is otherwise required to successfully use `cargo-leptos`.

An example where this is especially useful is a deploy pipeline for an AWS Lambda function. The server binary destined for the function must, among other things, target a particular architecture. [`cargo-lambda`](https://github.com/cargo-lambda/cargo-lambda) makes this very easy to achieve. Of course, the lib target and associated files do not require anything different, so they can be built with `cargo-leptos` as normal. Here's what this might look like currently:

```
cargo leptos build --release
cargo lambda build --no-default-features --features=ssr --release
cargo lambda deploy --include target/site --enable-function-url
```

In this sequence, the bin target is built twice. With a `--bin-skip` flag, it would only be built once:

```
cargo leptos build --release --bin-skip
cargo lambda build --no-default-features --features=ssr --release
cargo lambda deploy --include target/site --enable-function-url
```

## Implementation

My implementation is rather quick and dirty. I wanted to avoid adding a flag to all the subcommands when it only makes sense for `build`. Duplicating a large struct only to change one member is not ideal though. If there's a better approach for this feature, I'm all ears! I also totally understand if this feature isn't all that appropriate at the end of the day.

It may also make more sense to provide mutually exclusive flags (or an option of some sort) so either target can be skipped.

### DIFF
--- a/src/command/build.rs
+++ b/src/command/build.rs
@@ -11,11 +11,11 @@ use crate::{
     },
 };
 
-pub async fn build_all(conf: &Config) -> Result<()> {
+pub async fn build_all(conf: &Config, bin_skip: bool) -> Result<()> {
     let mut first_failed_project = None;
 
     for proj in &conf.projects {
-        if !build_proj(proj).await? && first_failed_project.is_none() {
+        if !build_proj(proj, bin_skip).await? && first_failed_project.is_none() {
             first_failed_project = Some(proj);
         }
     }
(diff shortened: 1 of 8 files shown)


luxalpa:
I think instead of skipping targets it would be nicer if we could be listing targets, like `build server`, `build client`, etc

benwis:
This is probably too old to actually merge, so I'll close this one. A couple of people have expressed interest in flags to only build the bin or the lib, so a new PR is always welcome

PR #214, opened 2023-11-13
document the environment variable to override the hardocded versions of the external tools

crapStone:


benwis:
Looks good. Thanks for the PR!

PR #218, opened 2023-11-16
Fix error with path unbase for tmp_dir when `CARGO_TARGET_DIR` is set

### LINKED ISSUE #215: "Could not remove base" error with `CARGO_TARGET_DIR`
`cargo leptos watch` fails when `CARGO_TARGET_DIR` is set.  I tried recent master (b3a556ae68cdb303b67d5919e54017c427f6983d)

1. `mkdir /tmp/test`
2. `cd /tmp/test`
3. `cargo leptos new --name project --git https://github.com/leptos-rs/start`
4. `cd /tmp/test/project`
5. `CARGO_TARGET_DIR=/tmp/test/.cache cargo leptos watch`

Results in

```
thread 'main' panicked at src/config/project.rs:217:14:
called `Result::unwrap()` on an `Err` value: Could not remove base "/tmp/test/project" from "/tmp/test/.cache/tmp"
```

Error happens at https://github.com/leptos-rs/cargo-leptos/blob/b3a556ae68cdb303b67d5919e54017c427f6983d/src/config/project.rs#L213-L217

Maneren:
This should fix #215. I have tested it with my project that uses both tailwind and custom `CARGO_TARGET_DIR` and it works.

I see no reason for the unbase to be there, but I am not that familiar with the codebase, so someone please check me.


benwis:
As far as I can tell, this works with no negative effects. Merging

benwis:
Thanks for the work and the PR!

PR #225, opened 2023-12-10
Fix passing multiple cargo args via `--{bin,lib}-cargo-args`

aqrln:
This PR fixes passing multiple cargo args via `--bin-cargo-args` and
`--lib-cargo-args`. Previously those arguments would be merged into a
single argument containing spaces inside, leading to errors.

Discovered when trying to cross-compile in a wrong way (before figuring
out `LEPTOS_BIN_TARGET_TRIPLE`) and getting this error:

```
error: unexpected argument '--target x86_64-unknown-linux-gnu' found

  tip: a similar argument exists: '--target'
```

### DIFF
--- a/src/compile/tests.rs
+++ b/src/compile/tests.rs
@@ -136,3 +136,23 @@ fn test_workspace_project2() {
 
     assert_display_snapshot!(cargo, @"cargo build --package=project2 --lib --target-dir=/celPool/celData/Works/projects/cargo-leptos/examples/workspace/target/front --target=wasm32-unknown-unknown --no-default-features --features=hydrate");
 }
+
+#[test]
+fn test_extra_cargo_args() {
+    let cli = Opts {
+        lib_cargo_args: Some(vec!["-j".into(), "8".into()]),
+        bin_cargo_args: Some(vec!["-j".into(), "16".into()]),
+        ..dev_opts()
+    };
+    let conf = Config::test_load(cli, "examples", "examples/project/Cargo.toml", true);
+
+    let mut command = Command::new("cargo");
+    let (_, cargo) = build_cargo_server_cmd("build", &conf.projects[0], &mut command);
+
+    assert_display_snapshot!(cargo, @"cargo build --package=example --bin=example --no-default-features --features=ssr -j 16");
+
+    let mut command = Command::new("cargo");
+    let (_, cargo) = build_cargo_front_cmd("build", true, &conf.projects[0], &mut command);
+
+    assert_display_snapshot!(cargo, @"cargo build --package=example --lib --target-dir=/celPool/celData/Works/projects/cargo-leptos/examples/project/target/front --target=wasm32-unknown-unknown --no-default-features --features=hydrate -j 8");
+}
--- a/src/compile/front.rs
+++ b/src/compile/front.rs
@@ -84,9 +84,7 @@ pub fn build_cargo_front_cmd(
 
     // Add cargo flags to cargo command
     if let Some(cargo_args) = &proj.lib.cargo_args {
-        if !cargo_args.is_empty() {
-            args.push(format!("{}", cargo_args.join(" ")))
-        }
+        args.extend_from_slice(cargo_args);
     }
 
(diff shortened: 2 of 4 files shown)


src/compile/tests.rs
```diff
@@ -136,3 +136,23 @@ fn test_workspace_project2() {
...
+        lib_cargo_args: Some(vec!["-j".into(), "8".into()]),
+        bin_cargo_args: Some(vec!["-j".into(), "16".into()]),
+        ..dev_opts()
+    };
+    let conf = Config::test_load(cli, "examples", "examples/project/Cargo.toml", true);
+
+    let mut command = Command::new("cargo");
+    let (_, cargo) = build_cargo_server_cmd("build", &conf.projects[0], &mut command);
+
+    assert_display_snapshot!(cargo, @"cargo build --package=example --bin=example --no-default-features --features=ssr -j 16");
```
aqrln, line 152: So, this test is still probably somewhat useful because the functionality wasn't tested before at all, but it... doesn't actually work as a regression test for this issue. The problem is joining everything to a single string for the snapshot masks the problem right now.

One way to fix this would be to wrap arguments that contain spaces into quotes when building this string, another would be to make `build_cargo_server_cmd` and `build_cargo_front_cmd` return a `Vec<String>` instead of a `String` as the second element in the tuple, if that's practical.

Should I do any of these in this PR?

aqrln: Actually I just went ahead and made the rendered cargo command wrap arguments containing spaces with quotes, see the second commit.

Now, if the bug is re-introduced, the snapshots will fail:

```
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Snapshot Summary ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Snapshot: extra_cargo_args
Source: src/compile/tests.rs:152
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Expression: cargo
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
-old snapshot
+new results
────────────┬─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
    0       │-cargo build --package=example --bin=example --no-default-features --features=ssr -j 16
          0 │+cargo build --package=example --bin=example --no-default-features --features=ssr '-j 16'
────────────┴─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
To update snapshots run `cargo insta review`
Stopped on the first failure. Run `cargo insta test` to run all snapshots.
```

It will also make the user facing output more correct in some cases since this also seems to be used in logs.

benwis:
That's some nice work, thanks for making those commands better. Admittedly when I wrote it I only really used it for --timin

aqrln:
These snapshots are failing because of absolute paths in other PRs as well: I thought that the hardcoded path corresponds to the path on CI but apparently not.

benwis:
> These snapshots are failing because of absolute paths in other PRs as well: I thought that the hardcoded path corresponds to the path on CI but apparently not. Do I need to do anything with it in this PR? I can submit a separate patch that deals with it if you'd like.

I really think we need to chuck these snapshots, no  I'm just going to merge this in. If you'd like to PR those tests out or fix it, I'd be thankful

PR #227, opened 2023-12-10
Add `.direnv/` to `.gitignore`

aqrln:
Add `.direnv` directory created by [`nix-direnv`](https://github.com/nix-community/nix-direnv) to `.gitignore`, since both direnv and nix are used in the project.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,5 @@ playwright/.cache/
 .DS_Store
 
 .idea/
+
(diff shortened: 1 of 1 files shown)


aqrln:
Probably either other folks are using the built-in simple `use flake` implementation in direnv rather than faster but more complex nix-direnv, or have `.direnv` in a global gitignore.

Issue #224, opened 2023-12-08
`bin.exe_file` on `BinPackage` does not consider cross-compilation


chrisp60:
https://github.com/leptos-rs/cargo-leptos/blob/170d9a09a22c68cf349514a09f1962185fc93333/src/config/bin_package.rs#L87C2-L87C3

https://github.com/leptos-rs/cargo-leptos/blob/170d9a09a22c68cf349514a09f1962185fc93333/src/service/site.rs#L117C6-L117C6

I develop on wsl and my application is compiled for windows for some use cases. Get the below error as its not looking for `kolibri.exe` but instead `kolibri`. Given, everything still compiles without issue. Seems to just be the hashing that fails.

I might have some time to fix this up, but can't say for sure.

```sh
Error: at `/home/[omitted]/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cargo-leptos-0.2.5/src/compile/server.rs:39:22`

Caused by:
    0: at `/home/[omitted]/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cargo-leptos-0.2.5/src/service/site.rs:117:44`
    1: Could not read "target/x86_64-pc-windows-gnu/release/kolibri" at `/home/[omitted]/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cargo-leptos-0.2.5/src/ext/fs.rs:44:10`
    2: No such file or directory (os error 2)
```

chrisp60:
After digging into some of the thrills of cross compilation I would just ask if it's even something `cargo_leptos` would want to support? There may be too many edge cases to make a commitment beyond building for the arch that we are already on.

benwis:
I could swear someone already did cross compilation using cargo cross and changing the bin-cargo-command arg

chrisp60:
Oh, apparently there was discussion on discord that pointed to an undocumented arg to override the cargo command, https://github.com/leptos-rs/cargo-leptos/blob/170d9a09a22c68cf349514a09f1962185fc93333/src/config/project.rs#L173C1-L174C43

They pointed it to cross. Which works fine. So maybe just best to document it.

benwis:
Closing this as #228 was merged

PR #228, opened 2023-12-21
Doc `bin-cargo-command` in `README.me`

chrisp60:
As discussed in #224 (better late than never I hope), `bin-cargo-command` is now included in the readme.

Secondly, I noticed two other lines were mixed between snake and kebab-case. Went ahead and corrected those.


benwis:
LGTM, thanks for submitting the PR. Happy holidays!

PR #244, opened 2024-01-24
Added dockerfile and workflow to push to dockerhub

rakshith-ravi:


rakshith-ravi:
I have the dockerhub credentials. Need access to the secrets to add it

rakshith-ravi:
What is `cargo dist` and why does it not like me? 😆

benwis:
> What is `cargo dist` and why does it not like me? 😆

cargo-dist is responsible for packaging all of our binaries for releases

Issue #245, opened 2024-01-29
Pass command line arguments to bin in `cargo leptos watch`?


jmjoy:
For example, and add arg `--bin-cmd-args` for `cargo leptos watch`, or pass cmd args after `--`.

PR #246, opened 2024-01-31
Add bin args for `watch` and `serve` commands

jmjoy:
For #245 

So we can pass command line arguments to bin, such as `cargo leptos watch -v -- --foo`.


zakstucke:
@benwis for me this is useful to pass server config args processed with clap, I often find them more ergonomic than env vars! 

Would be great if this could get merged :)

Issue #247, opened 2024-01-31
Incorrect help message for very verbose option


dav-wolff:
Running `cargo leptos build --help` says to use the `--vv` option for very verbose output:
```
-v...
        Verbosity (none: info, errors & warnings, -v: verbose, --vv: very verbose)
```

However, running `cargo leptos build --vv` produces the following error:
```
+ command cargo leptos build --vv
error: unexpected argument '--vv' found

Usage: cargo-leptos build [OPTIONS]

For more information, try '--help'.
```

It should say to use `-vv` instead of `--vv`. The same applies to test, end-to-end, serve, and watch.

gbj:
Thanks. Would you be interested in making a PR to fix it?

dav-wolff:
Yes, will do

PR #248, opened 2024-01-31
Fix help message for verbosity option

dav-wolff:
Fixes issue #247 

Tell the user to use the `-vv` flag for very verbose, instead of the incorrect `--vv`.


benwis:
LGTM. Thank you!

Issue #217, opened 2023-11-16
Remove `separate-front-target-dir` altogether in 0.3.x


rakshith-ravi:
Keeping this as a reminder here (we don't have milestones on GH do we?).

Once we reach 0.3, we should remove the depreciation notice (introduced in #216) and remove the option altogether

PR #251, opened 2024-02-08
Explain what has become of separate-front-target-dir

g2p:
The change in #216 did not explain enough and contradicted the way the option was documented in the README.

See #203 and #217.


benwis:
Thank you for contributing

PR #261, opened 2024-03-03
LEPTOS_WATCH=true instead of ON

maccesch:


benwis:
I think you probably want to rebase this to have maybe 1 commit. 😃

PR #267, opened 2024-03-15
Fixed Output Bug

nvim-ftw:
Fixed an output bug where Dart Sass CSS compilation would display info saying it was for Tailwind.


benwis:
Thanks for noticing the issue and writing a PR!

PR #268, opened 2024-03-18
Fix cargo args

emil-dietrich:
`*-cargo-args` was expected to be a `String` even tough the docs shows them as arrays.

The empty cli flag for `lib-cargo-args` no longer overrides the one provided in the config.
I have not further investigated `bin-cargo-args`.


benwis:
I think this makes sense, thanks for catching it!

PR #285, opened 2024-05-21
Feat: Minify javascript

Th3Whit3Wolf:
Adds [swc](https://github.com/swc-project/swc) (Speedy Web Compiler) to minify javascript.

I have seen around 50% reduction in javascript filesize.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,16 @@
 version = 3
 
 [[package]]
+name = "Inflector"
+version = "0.11.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fe438c63458706e03479442743baae6c88256498e6431708f6dfc520a26515d3"
+dependencies = [
+ "lazy_static",
+ "regex",
+]
+
+[[package]]
 name = "addr2line"
 version = "0.21.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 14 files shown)


gbj:
Overall: This looks good to me, and good idea.

If I'm remembering the structure of cargo-leptos correctly, this would run every time the JS/WASM are rebuilt? I would suggest that this should either have a way to opt out, or should be opt-in for release builds or something, for a few reasons:
1. Compile speed: There's some non-zero overhead to minifying the JS which makes recompilation marginally slower, but there's little benefit to having minified JS in dev mode
2. Debugging: In fact there's some drawback to always minifying and mangling, which is that it makes your JS always unreadable
3. Bugs: If there's some behavior in either SWC or wasm-bindgen that breaks minified wasm-bindgen JS, cargo-leptos becomes unusable
4. Other minifiers: People may have existing setups with other JS build tools

So yeah, I'd say adding a configuration variable (even default to `minify = true` or whatever) and checking that before minifying would be my preferred approach.

Th3Whit3Wolf:
Okay I can fix that. Would you prefer that to variable to be named `minify` or `js-minify`? I know that there is minify configuration for CSS but it does seem like it could be desirable in the future if there was ever any issue with lightningcss and `js-minify`, to me, has a clearer expected behavior.

benwis:
I'd say js-minify would be better, for clarity's sake

Th3Whit3Wolf:
Changes implemented:

* Added `js_minify` field to `Project` that defaults to true and can be set via:
  * `--js-minify` (uses [BoolishValueParser](https://docs.rs/clap/latest/clap/builder/struct.BoolishValueParser.html))
  * `js-minify` config option
  * `LEPTOS_JS_MINIFY` environment variable
* Minifies javascript only when in release and `js_minify` has not been set to false

Th3Whit3Wolf:
Noticed an issue when embedding compressed assets with `hash_files = true`. Fixed it in last commit. This PR is ready for review.

benwis (approved):
Everything here looks good, nice work. I haven't tried to use SWC before, we'll have to run it through the gauntlet in git

PR #289, opened 2024-06-05
Run tailwindcss command instead of tailwind

SleeplessOne1917:
Small change to the CLI command used to run tailwind. I've run into a difficult to debug error that I suspect is caused by what is there before this change. Regardless, [the page on Tailwind CLI](https://tailwindcss.com/docs/installation) uses `tailwindcss` as the command as opposed to `tailwind`.


benwis:
That's a very odd one, hopefully this fixes it for you!

PR #292, opened 2024-06-12
Minor: Bumped gloo-net to version 0.5 ( was a mix of 0.2 and 0.4 ) 

martinfrances107:
Maintenance patch. #274 is too complex to review and reason with/about

So I am breaking it up into a series of micro-step, each of which I know is honest and true.

#291 was the first -- giving some TLC to gloo-net is next.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,11 +4,11 @@ version = 3
 
 [[package]]
 name = "addr2line"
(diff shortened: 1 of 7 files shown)


martinfrances107:
Ok -- so exposes the complex difficult to maintain relations between sub-modules

PR #293, opened 2024-06-12
Minor: examples - Bump assorted leptos sub-modules up to version 0.6.'

martinfrances107:
As of leptos 0.5 there is a major change:  the cx variable has been removed so there are lots of changes of the form below

```rustlang
-pub fn App(cx: Scope) -> impl IntoView {
-    provide_meta_context(cx);
+pub fn App() -> impl IntoView {

     view! {
-        cx,
         <div>
```

NB This is all part of a mini project to ship wreak  #274... break the PR into part that are easy to reivew.

### DIFF
--- a/examples/workspace/Cargo.lock
+++ b/examples/workspace/Cargo.lock
@@ -52,7 +52,7 @@ dependencies = [
  "actix-rt",
  "actix-service",
  "actix-utils",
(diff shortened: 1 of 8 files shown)


benwis:
I do find myself wondering we should remove these examples, they seem redundant. But this is a nice fix

PR #294, opened 2024-06-12
Minor: Bump version numbers for Itertools and Brotli.

martinfrances107:
So I am splitting up #274 into small parts.

Eventually that MEGA patch will be so small that the difficult to reason with aspect of the PR becomes workable.


benwis:
LGTM!

PR #295, opened 2024-06-13
Minor:  zip bump version number.

martinfrances107:
Over in #274 , I am trying to simplify an mega PR.

This is the next round of crate updates that I think are safe to split off 
so I can reason with a subtle bug.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4705,14 +4705,15 @@ dependencies = [
 
 [[package]]
 name = "which"
-version = "6.0.1"
+version = "5.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


martinfrances107:
Ah that is an interesting data point... 

**Just updating lightningcss and zip as a pair causes problems ...**

By way of discussion ... relying on a alpha level crates in production code 
may just have biten us in the rear area.

lightningcss = { version = "1.0.0-alpha.47", features = ["browserslist"] }

So I am backing out changes to lightningcss

martinfrances107:
Partially bumping to 1.0 shows that the breaking changes happen early.

martinfrances107:
No longer relatable to "HEAD"

PR #296, opened 2024-06-14
Minor: Bump crate versions for which and lightningcss

martinfrances107:
Over in #274 I am wresting with a complex coupling between crates

The way do reason with the problem is to strip off crates updates into small managable PRs 
These should be a breeze to review... and incidently reducing the complexity of the mega patch...

Here I am pairing up updated for the crates "which" and lightningcss


benwis:
LGTM

PR #298, opened 2024-06-14
Minor: Bump Reqwest to 0.12.

martinfrances107:
Over in #274  I am trying to reduce the complexity of a MEGA PR 

Here I am splitting off changes to the reqwest crate to see if it unwinds a 
suble unwanted coupling between crates.

Currently is looks like the "notify" and "zip" crates are entangled in a way that makes me grumpy.


benwis:
Thanks!

PR #301, opened 2024-06-17
Minor: Bumped swc to 0.287.

martinfrances107:


martinfrances107:
Works on Linux but fails on Windows.

notfiy mentioned .... hmmm where have I seen that before ( #274 ) 

```
error[E0308]: mismatched types
    --> C:\Users\runneradmin/.cargo\registry\src\index.crates.io-6f17d22bba15001f\notify-4.0.17\src\windows.rs:282:13
     |
280  |         let ret = winbase::ReadDirectoryChangesW(
     |                   ------------------------------ arguments to this function are incorrect
281  |             handle,
282  |             req_buf,
     |             ^^^^^^^ expected `winapi::ctypes::c_void`, found `libc::c_void`
     |
     = note: `libc::c_void` and `winapi::ctypes::c_void` have similar names, but are actually distinct types
```

benwis:
Putting a not here that this is failing because of winapi, which needs a feature flag it doesn't seem to have. Not sure where it's doing this

PR #304, opened 2024-06-17
fix: edge case for path_dependency == workspace_root (Fixes #303)

### LINKED ISSUE #303: Doesn't watch if the path dependency is the workspace root
Leptos-project: `my-library/crates/website`
workspace root: `my-library`

In the above case, if the leptos project has a path dependency on the workspace root, as in `path = "../.."` it will not be able to watch the files in the workspace root's src folder. This is because in this case `unbase` will give a `"."` path buf, which is then joined to `src`, yielding `./src` as the path, which then fails the path prefix check for `src/lib.rs`

https://github.com/leptos-rs/cargo-leptos/blob/d36a1870452fceab2365df6ae16677b6573b69a3/src/ext/cargo.rs#L104-L106

luxalpa:
I'm unfamiliar with the camino crate so I'm not sure if maybe there's a more elegant way to write this.

### DIFF
--- a/src/ext/cargo.rs
+++ b/src/ext/cargo.rs
@@ -101,9 +101,13 @@ impl MetadataExt for Metadata {
         self.path_dependencies(id)
             .iter()
             .map(|p| {
-                p.unbase(root)
-                    .unwrap_or_else(|_| p.to_path_buf())
-                    .join("src")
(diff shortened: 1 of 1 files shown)


benwis (approved):
LGTM

Issue #241, opened 2024-01-21
Provide way to disable minification as Lightning CSS can break css when minified


DanielJoyce:
By default, release mode enables Lightning minification, but there are outstanding bugs where the transforms lightning uses can break CSS intention.

We might want a way to disable minification in release mode, or at least, only minimize whitespace and nothing else.

DanielJoyce:
Another option may be to look at swc-css-minifier which is not based on lightning.

benwis:
Might be worth looking at, depending on the activity on lightningcss and that package

DanielJoyce:
Lightning has some issues with order of properties and I personally think the compression is over-eager, clobbering overrides. 

SWC minifier is slower by about 4x, but still very fast. 

I think a disable-minification switch is a quick first one, followed by maybe a option to use swc-css behind a feature flag.

benwis:
Sounds good to me :)

bematteo:
Having the same issue (I think?).
I tried putting `@import "daisyui.full.min";` in my main.scss and the resulting style seems to be broken.
Works well if including from CDN.

karpfediem:
So i made a PR which allows setting the minify option to false for lightningcss.
Yet it did not fix my issue with DaisyUI appearing broken.
Here are some insights by debugging:
- Tailwind gets called and generates tailwind.css inside the target/tmp/ directory. Including this file directly, DaisyUI looks perfectly fine.
- With my new option i can skip the call of `stylesheet.minify(minify_options)`
- I am setting PrinterOptions with minify: false, and then call `stylesheet.to_css(options)`
- Something about lightningcss parsing the CSS and then handing it back out is breaking some things about DaisyUI

karpfediem:
Here is the output of `diff test.css target/site/pkg/site.S9Dnrx_tfOQrVk20EahLIw.css`
https://pastebin.com/knX83mha

tailwind.css https://pastebin.com/PbMk1Jyn
generated site.S9Dnrx_tfOQrVk20EahLIw.css https://pastebin.com/1ht1a3d9

benwis:
I can't say I have an answer for this issue, but you could not use cargo-leptos to process your css. Putting it in your assets folder would make it available.

CodingFlow:
Any chance disabling css minification in release builds could still be implemented? I have a use case where it would be beneficial.

PR #306, opened 2024-06-22
Draft: Conditionally minify css

karpfediem:
Adds `css_minify` option in line with https://github.com/leptos-rs/cargo-leptos/commit/7b316cfc7ae7264084a7c737eec99e0c653f2db5 (see also https://github.com/leptos-rs/cargo-leptos/pull/285)

I am using DaisyUI and currently using lightningcss seems to break it (https://github.com/leptos-rs/cargo-leptos/issues/241)

I saw the `js_minify` option that was just added, so i think a `css_minify` should also exist which works in the same way.

This is logically complete, and working. Yet setting minify to false for lightningcss did not solve my issues with the faulty processing.
I put this up as a Draft for now to allow an easier overview/discussion.


karpfediem:
serves no purpose for me personally
closed to prevent clutter

PR #316, opened 2024-07-15
feat: add support for graceful exits

rupansh:
This adds a new flag `--graceful-exit`, which sends `SIGTERM` signal to the server process on unix hosts and `Ctrl+C` on windows.

Note that for windows, I only checked the compilation. so testing on windows would be greatly appreciated.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -479,6 +479,7 @@ dependencies = [
  "lightningcss",
  "log",
  "md-5",
+ "nix",
  "notify",
  "pathdiff",
  "reqwest",
@@ -494,6 +495,7 @@ dependencies = [
  "tokio",
  "wasm-bindgen-cli-support",
  "which",
+ "windows",
  "zip",
 ]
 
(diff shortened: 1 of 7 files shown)


rupansh:
I fired up a windows VM for testing and found that CTRL+C doesn't work. Applications must handle ctrl+break on windows for graceful exits.
Note that on windows, CTRL+C does happen gracefully on a specific condition, when the parent process receives CTRL+C as well.

here's the termination future I passed to `with_graceful_shutdown` in axum:

```rust
    let terminate = {
        use tokio::signal;

        let ctrl_c = async {
            signal::ctrl_c()
                .await
                .expect("failed to install Ctrl+C handler");
        };

        #[cfg(unix)]
        let terminate = async {
            signal::unix::signal(signal::unix::SignalKind::terminate())
                .expect("failed to install signal handler")
                .recv()
                .await;
        };

        #[cfg(windows)]
        let terminate = async {
             signal::windows::ctrl_break()
                 .expect("failed to install Ctrl+Break handler")
		     .recv()
                 .await;
	  };

        async {
            tokio::select! {
                _ = ctrl_c => {},
                _ = terminate => {},
            }
            logging::log!("stopping...");
        }
    };
    ```

benwis:
This has some conflicts and is quite old now, if you have use for this feel free to fix those and comment here, and I'll see about merging this

PR #318, opened 2024-07-30
Permit explicit target opt-in

CorvusPrudens:
This PR supersedes #212 with a more general solution. The utility is explained there.

Rather than providing a single `--bin-skip` flag, the build command now has two additional flags `--bin` and `--lib`. These provide explicit opt-ins for their respective targets.

The default behavior is to build both. If an explicit opt-in is provided for one target, the other target will be excluded unless it also receives an opt-in.

This behavior might be a bit confusing. If there's a better and simpler approach, for example simply providing a boolean value for each, that should be an easy change.

If additional testing is required, please let me know! I simply adjusted the tests assuming no flags were provided.


benwis:
@CorvusPrudens Sorry about the delayed review, if you can fix the conflicts I'd be happy to merge this!

CorvusPrudens:
Looking at the CLI flags again, I think it may make more sense to take a boolean value.

For example:
```
cargo leptos build --bin false
```

There are already precedents for boolean options right in the build subcommand, and it makes the functionality more clear.

I should be able to implement this very soon. It's really not much different from how it works already.

benwis:
I think I'd do a --bin-only or a --lib-only flag, that way it's clear what they do

benwis:
I'm going to close this PR as stale, I still would like this feature, so if you feel like fixing the conflicts, feel free to reopen

PR #319, opened 2024-08-02
Change the default wasm-opt settings to optimize for size

Innominus:
Currently wasm-opt is set to compile for speed, but if we are setting a default for wasm-opt for users than I believe most people would want it optimizing for size instead.

### DIFF
--- a/src/compile/front.rs
+++ b/src/compile/front.rs
@@ -173,7 +173,7 @@ async fn optimize(
 ) -> Result<CommandResult<()>> {
     let wasm_opt = Exe::WasmOpt.get().await.dot()?;
 
-    let args = [file.as_str(), "-Os", "-o", file.as_str()];
+    let args = [file.as_str(), "-Oz", "-o", file.as_str()];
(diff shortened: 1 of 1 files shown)


gbj:
This is good — but just a note that `s` and `z` are both optimizations for size, the difference being that [`z` turns off loop vectorization](https://doc.rust-lang.org/cargo/reference/profiles.html#opt-level). Per Rust docs:
> It is recommended to experiment with different levels to find the right balance for your project. There may be surprising results, such as level `3` being slower than `2`, or the `"s"` and `"z"` levels not being necessarily smaller.

Ultimately defaulting to `z` and providing a user option to configure it is probably best.

Innominus:
Apologies, yes both are optimised for size with one loop unrolling. Unfortunately I don't think I'm at a level with Rust capable of implementing user options to configure. Defaulting to `z` for now makes sense so I can move off my branch (need minimal size), and if this still isn't implemented with user configuration in the future I'd be happy to try take a stab at it.

PR #321, opened 2024-08-05
fix: report the error on `compress_static_files` fail

sabify:


sabify:
The CI is green now after the rebase.

Issue #320, opened 2024-08-03
workspace example is broken with LEPTOS_HASH_FILES=true


alexichepura:
I have setup similar to workspace example. Building it with LEPTOS_HASH_FILES gives an error:
```sh
cd examples/workspace
LEPTOS_HASH_FILES=true cargo leptos build --release --project project2
Error: No such file or directory (os error 2)
```
I found that target directory is not correct, should be examples/workspace/target.
```sh
LEPTOS_HASH_FILES=true cargo leptos build --release --project project2
# [src/compile/hash.rs:22:5] &proj.hash_file.abs = "/Users/alexi/w/cargo-leptos/examples/workspace/project2/target/release/hash.txt"
```
```sh
LEPTOS_HASH_FILES=true LEPTOS_BIN_TARGET_TRIPLE=x86_64-unknown-linux-gnu cargo leptos build --release --project project2
# [src/compile/hash.rs:22:5] &proj.hash_file.abs = "/Users/alexi/w/cargo-leptos/examples/workspace/project2/target/x86_64-unknown-linux-gnu/release/hash.txt"
```
I think it's after this PR with tweaks for hash_file #282, that needs to be improved for workspace support.
I'll try to make the PR in the next few days as i need this changes anyway.

benwis:
Fixed in 2.19

PR #329, opened 2024-08-12
fix for hash_file to work in workspace and target

alexichepura:
Fix for #320, partially reverts #282.
I tested it in my project and it works, 
with and without LEPTOS_BIN_TARGET_TRIPLE, with and without LEPTOS_BIN_TARGET_DIR

I could not yet build the workspace example, it throws an error for "time" crate - error[E0282]: type annotations needed for `Box<_>`

I noticed that it will fail with error if there is no target directory yet, like x86_64-unknown-linux-gnu/release, so I had to create it manually for now. I'll recheck later and will create an issue if need.
Error: No such file or directory (os error 2)

### DIFF
--- a/src/config/hash_file.rs
+++ b/src/config/hash_file.rs
@@ -8,13 +8,13 @@ pub struct HashFile {
 }
 
 impl HashFile {
-    pub fn new(bin: &BinPackage, rel: Option<&Utf8PathBuf>) -> Self {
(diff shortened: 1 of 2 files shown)


benwis:
Try running `cargo update` on the repo to fix the time issue

benwis:
I screwed up git and merged this seperately. Thanks for the PR!

PR #345, opened 2024-08-18
hash.txt path correction

siqpush:
hash.txt file was being looked for at the bin exec path but lives at the workspace root level


benwis:
Thank you!

Issue #347, opened 2024-08-20
`hash_files = true` doesn't work anymore


maccesch:
It builds the files with the hashes in the name but then requests the files without hash.

If suspect that #345 broke non-workspace builds. Haven't tested it yet, though.

In my project I'm not setting env vars but have it configured in Cargo.toml

benwis:
Hmm, actually reading the files is the responsibility of Leptos proper, as long as we're still setting the correct env vars. I'll have to take a look

maccesch:
IIRC when leptos proper doesn't find the hash.txt it simply defaults to loading without hashes. So maybe that's the case here?

benwis:
> IIRC when leptos proper doesn't find the hash.txt it simply defaults to loading without hashes. So maybe that's the case here?

Yeah, I think I've fixed that. Can you try the latest git main and lmk?

maccesch:
I can confirm the latest git main works for my project. I suspect it won't anymore for workspace projects? I can't test that right now.

lajtomekadimon:
I'm using workspaces, and when I try to compile WASM, I get a generic "no such file or directory" error:

```
root@d01bb3cc9efd:/srv/myproject# cargo leptos watch 
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.18s
       Cargo finished cargo build --package=myproject_server_hydration --lib --target-dir=/srv/myproject/target/front --target=wasm32-unknown-unknown --no-default-features
       Front compiling WASM
Error: No such file or directory (os error 2)
```

I didn't change anything from my project. Before upgrading to the last release, everything worked fine.

Leptos version: 0.6.14
Cargo Leptos version: lastest git main.

benwis:
> I'm using workspaces, and when I try to compile WASM, I get a generic "no such file or directory" error:
> 
> ```
> root@d01bb3cc9efd:/srv/myproject# cargo leptos watch 
>     Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.18s
>        Cargo finished cargo build --package=myproject_server_hydration --lib --target-dir=/srv/myproject/target/front --target=wasm32-unknown-unknown --no-default-features
>        Front compiling WASM
> Error: No such file or directory (os error 2)
> ```
> 
> I didn't change anything from my project. Before upgrading to the last release, everything worked fine.
> 
> Leptos version: 0.6.14 Cargo Leptos version: lastest git main.

[...]

PR #353, opened 2024-08-21
create parent dir for hashfile if not exists

maccesch:
Relates to #347

This makes sure that hashing works after `cargo clean`.


benwis:
Seems reasonable to me. Thank you!

PR #360, opened 2024-09-02
feat: implemented workaround to enable short form instantiation of all offical leptos-rs templates

diversable:
NB: this is a tiny change that doesn’t modify any logic, just makes leptos templates a little quicker to get up & running.

Also included a new style of short-form template instantiation for official templates, without the `leptos-rs/` prefix:
A template can now alternatively be instantiated like 
`cargo leptos new —git start-axum`, 
`cargo leptos new —git start-actix`, or 
`cargo leptos new —git start-spin`;
the old template instantiation behaviour still exists (and now exists for all templates), so eg.
`cargo leptos new —git leptos-rs/start-aws` will work as well.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -457,7 +457,7 @@ dependencies = [
 
 [[package]]
 name = "cargo-leptos"
-version = "0.2.19"
+version = "0.2.20"
 dependencies = [
  "ansi_term",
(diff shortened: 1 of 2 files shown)


benwis:
This a neat little tweak, thanks for the PR

PR #372, opened 2024-10-01
Chore: Updating lots of crates

martinfrances107:
I did not bump all the modules reported by "outdated" 

There is a long standing sore #274 ... which is in limbo 
Where I am having trouble bumping "notify" -- I don't have a window build to debug the issues.

274 spawned lots of minor patches fixing unrelated issues... time moves on and this patch 
is just nibbling at the edges fixing more unrelated issues.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14,18 +14,18 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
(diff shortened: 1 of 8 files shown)


martinfrances107:
Thank you

Issue #376, opened 2024-10-16
Hash files on wasm snippet breaks wasm imports


DennisJensen95:
Inline{num}.js with hash breaks the imports in webassembly package in the frontend. 

So the case is the following where i have my site directory:

```bash
tree site/pkg 
site/pkg
├── digitalt-haandvaerk.HuD81diBfNuG42TNaU25xQ.js
├── digitalt-haandvaerk.i9AokiMJC7DE9hQ8undZAA.css
├── digitalt-haandvaerk.vLtC8wK4np8rkwpuF5QdJA.wasm
└── snippets
    └── openlayer-rs-c384bd82f6cf7534
        └── inline0.g3NfIIkxtWHXCd96hHqZxg.js
```

So i am importing my own module like so

```toml
openlayer-rs = { path = "../../lib/openlayer-rs" }
```

Now it works just fine when i do not run my build with `LEPTOS_HASH_FILES=true`.

But when i do put on the hash my webassembly code is looking for the wrong inline0.js file. The error is the following:

```log
localhost/:1 Uncaught (in promise) TypeError: WebAssembly.instantiate(): Import #44 "./snippets/openlayer-rs-c384bd82f6cf7534/inline0.js": module is not an object or function
```

benwis:
Resolved

PR #377, opened 2024-10-16
fix(Hash-376): Hash should not be applied on inline.js

DennisJensen95:
Changes:
- Do not hash inline.js files as it will break the imports when using the webassembly in the frontend
- When hashing files, if the file is within snippets is called inline and has an extension of .js then do not hash it

### DIFF
--- a/src/compile/hash.rs
+++ b/src/compile/hash.rs
@@ -77,6 +77,25 @@ fn compute_front_file_hashes(proj: &Project) -> Result<HashMap<Utf8PathBuf, Stri
                         }
                     }
 
+                    // Check if the path contains snippets and also if it
+                    // contains inline{}.js. We do not want to hash these files
+                    // as the webassembly will look for an unhashed version of
+                    // the .js file. The folder though can be hashed.
+                    if let Some(path_str) = path.to_str() {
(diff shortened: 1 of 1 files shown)


benwis (approved):
Thanks for the PR, I'm curious how you're getting these inline js files

DennisJensen95:
I got the inline.js by having my own local library for binding openlayer js and using wasm-bindgen 0.2.93. Then it produces it.

DennisJensen95:
I tried to describe it in #376 🙂

DennisJensen95:
@benwis how do i merge this, is that only maintainers? 😊

benwis:
> @benwis how do i merge this, is that only maintainers? 😊

Yeah, only maintainers can merge. I suppose I'm trying to figure out if this is something that's general and should go in as is or whether it's somewhat specific to you, in which case we'd want a way to configure it

DennisJensen95:
> > @benwis how do i merge this, is that only maintainers? 😊
> 
> Yeah, only maintainers can merge. I suppose I'm trying to figure out if this is something that's general and should go in as is or whether it's somewhat specific to you, in which case we'd want a way to configure it

Totally fair, let me know if I can do anything 😊

Issue #400, opened 2024-11-29
bin-cargo-args from config file are ignored


collinpa:
bin-cargo-args from the config file are ignored and only applied from command line args.

Applying the example flags from the readme don't work for the bin target:

```
bin-cargo-args = ["--timings"]
```

They do work for the lib target.

Easiest to just point out the code. The lib target has the correct functionality

```rust
// src/config/lib_package.rs
        let cargo_args = cli
            .lib_cargo_args
            .clone()
            .or_else(|| config.lib_cargo_args.clone());

        Ok(Self {
            name,
            abs_dir,
            rel_dir,
            wasm_file,
            js_file,
            features,
            default_features: config.lib_default_features,
            output_name,
            src_paths: src_deps,
            front_target_path,
            profile,
            cargo_args,
        })

```

The bin target doesn't have this logic and only applies it from the cli.

```rust
// src/config/bin_package.rs
        log::debug!("BEFORE BIN {:?}", config.bin_cargo_command);
        Ok(Self {
            name,
            abs_dir,
            rel_dir,
            exe_file,
            target: target.name,
            features,
            default_features: config.bin_default_features,
            src_paths,
            profile,
            target_triple: config.bin_target_triple.clone(),
            target_dir: config.bin_target_dir.clone(),
            cargo_command: config.bin_cargo_command.clone(),
            cargo_args: cli.bin_cargo_args.clone(),
            bin_args: bin_args.map(ToOwned::to_owned),
        })

```

It seems like it could be fixed with the following change.

```rust
// src/config/bin_package.rs
        let cargo_args = cli
            .bin_cargo_args
            .clone()
            .or_else(|| config.bin_cargo_args.clone());

        log::debug!("BEFORE BIN {:?}", config.bin_cargo_command);
        Ok(Self {
            name,
            abs_dir,
            rel_dir,
            exe_file,
            target: target.name,
            features,
            default_features: config.bin_default_features,
            src_paths,
            profile,
            target_triple: config.bin_target_triple.clone(),
            target_dir: config.bin_target_dir.clone(),
            cargo_command: config.bin_cargo_command.clone(),
            cargo_args,
            bin_args: bin_args.map(ToOwned::to_owned),
        })

```

Note I haven't tested it

PR #401, opened 2024-11-29
fix: read ignored bin_cargo_args from config file

collinpa:
fixes https://github.com/leptos-rs/cargo-leptos/issues/400


benwis (approved):
LGTM :+1:

PR #404, opened 2024-12-04
fix: upgraded deps, simplified codebase and fixed logic for watch/notify/patch

sabify:


sabify:
What's changed:

- Updated and upgraded all dependencies.
- Simplified the watch, notify, patch and assets workflows that make it less error-prone, react to changes correctly and it's compatible with the latest `notify` library.
- Watcher now reacts correctly to the files' changes while running workflows, it will interrupt the workflow and rerun it to keep up with the changes.

src/signal/interrupt.rs
```diff
@@ -48,17 +48,12 @@ impl Interrupt {
 
     pub fn send(changes: &[Change]) {
         let mut ch = SOURCE_CHANGES.blocking_write();
-        let mut did_change = false;
         for change in changes {
-            did_change |= ch.add(change.clone());
+            ch.add(change.clone());
```
sabify, line 52: This change correctly notifies the watcher to terminate the running commands and rerun them.

sabify:
The failed CI may not be related to this PR. It happened after merging the main.

benwis:
Works well now that dist is updated

Issue #419, opened 2025-01-20
Cargo install cargo-leptos fails


BraveMole:
Cargo install cargo-leptos fails with the following error:

```
unresolved imports `swc_ecma_ast::fast_id`, `swc_ecma_ast::FastId`
 --> /home/bsouchu/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/swc_ecma_utils-6.0.1/src/ident.rs:3:20
  |
3 | use swc_ecma_ast::{fast_id, BindingIdent, FastId, Id, Ident};

```

Looks like swc has been recently updated and that is what is causing the error. In my compilation, swc v9.0.2 was used.

Note: cloned cargo leptos and built it, absolutly no problem. It only happens with cargo install

BraveMole:
Posted an issue on SWC : https://github.com/swc-project/swc/issues/9906

eagleflo:
swc moves pretty fast and is unlikely to fix issues with old versions; Donny has already closed the linked issue. I tried bumping swc to 10.0 and it seems to build & work fine.

blopker:
For anyone who just ran into this and want's a quick fix, use `--locked`:

```
cargo install cargo-leptos --locked
```

This uses the published Cargo.lock file instead of randomly updating dependencies. Maybe this should be mentioned on https://book.leptos.dev/ssr/21_cargo_leptos.html?

benwis:
@BraveMole should we pin to 10.0 as @eagleflo mentioned?

PR #420, opened 2025-01-20
Pin swc_ecma_ast to 5.0.1

BraveMole:
Fixes: https://github.com/leptos-rs/cargo-leptos/issues/419

Not the prettiest fix, but cargo install is not using Cargo.lock, so we need to pin swc_ecma_ast in Cargo.toml instead


benwis:
I think we're past this set of issues, if not feel free to reopen or comment here

Issue #358, opened 2024-09-01
cargo leptos buil / serve fails


holg:
cargo leptos build looks for the server binary always in 
target/debug
and then fails with:

```
Error: at `~/.cargo/git/checkouts/cargo-leptos-eb87bffcd2ca8f75/5dd5ec2/src/compile/server.rs:40:22`

Caused by:
    0: at `~/.cargo/git/checkouts/cargo-leptos-eb87bffcd2ca8f75/5dd5ec2/src/service/site.rs:117:44`
    1: Could not read "target/debug/test_leptos" at `/Users/htr/.cargo/git/checkouts/cargo-leptos-eb87bffcd2ca8f75/5dd5ec2/src/ext/fs.rs:44:10`
    2: No such file or directory (os error 2)
```
in fact everything builds well,
but the binary is located in target/TARGET_TRIPLE/debug.
In my case i can find the binary in target/aarch64-apple-darwin/
if i copy this into target/debug
it works, but sure this i not nice.

holg:
it still happens, when you do cross platform builds, so to check as well for the TARGET_TRIPLE, does not harm i guess?

PR #359, opened 2024-09-01
fix issue #358

holg:
cargo leptos build looks for the server binary always in
target/debug
and then fails with:

Error: at ~/.cargo/git/checkouts/cargo-leptos-eb87bffcd2ca8f75/5dd5ec2/src/compile/server.rs:40:22

```
Caused by:
    0: at `~/.cargo/git/checkouts/cargo-leptos-eb87bffcd2ca8f75/5dd5ec2/src/service/site.rs:117:44
    1: Could not read "target/debug/test_leptos" at `~/.cargo/git/checkouts/cargo-leptos-eb87bffcd2ca8f75/5dd5ec2/src/ext/fs.rs:44:10
    2: No such file or directory (os error 2)
```
For convienience i added the `current_platform = "0.2"` to the dependencies
So in bin_package i added to test for the file to exist, 
if not  use the TARGET_TRIPLE before debug:
    

    let mut test_file = file.join(profile.to_string())
                .join(name)
                .with_extension(file_ext);
            // Check if the file exists and if not, try to prepend target_triple
            // right now it mail fail to find target/debug/name
            // but the build is successful and in target/"target_triple"/debug/name
            // https://github.com/leptos-rs/cargo-leptos/issues/358
            if !test_file.exists(){
                test_file = Utf8PathBuf::from(format!(
                    "target/{}/{}/{}",
                    CURRENT_PLATFORM, profile.to_string(),test_file.file_name().unwrap()
                ));
            }
            test_file

PR #422, opened 2025-01-22
cargo leptos build fails, fixed issue for macos, which always builds into target/aarch64-apple-darwin

holg:
cargo leptos build fails, fixed issue for macos, which always builds into target/aarch64-apple-darwin
The issue here, as well as for cross-platform builds, is:
on macos the binary is alwasy put into target/aarch64[|x86_64]/
thus the binary later is not found.
I already made a fix, but it failed some tests, sorry!
Now i made the fix differently, to always add the target directly in the config.
The tests needed to be adapted to be more dynamic, as the target_triple is variable.
Have as well a look onto 
fix issue #358 #359
in which @gbj rightfully complained about the kind of the fix.
So it shall be OK now.


benwis:
@holg Can you fix the conflicts, and I'll review the changes?

holg:
> @holg Can you fix the conflicts, and I'll review the changes?
Sure, here you are

benwis:
The downside to this PR is that it changes the location of the output files for at least some of our users. Instead of being in `target/site` it's now in `target/{arch}/site`, which makes this a breaking change for most CI processes.

I'm also less sure if I noticed this behavior on my M1 mac, so I wonder if there's something more involved here

holg:
> The downside to this PR is that it changes the location of the output files for at least some of our users. Instead of being in `target/site` it's now in `target/{arch}/site`, which makes this a breaking change for most CI processes.
> 
> I'm also less sure if I noticed this behavior on my M1 mac, so I wonder if there's something more involved here

Hi Ben,
well this makes me wonder, ofc it is not intended to put the output files into target/{arch}/site,
i am looking to reproduce, as for me it is put into target/site...

PR #428, opened 2025-02-01
Tailwind v4 support

### LINKED ISSUE #417: Tailwindcss v0.4 support
Cargo leptos use `--config` option and provided config file to build tailwindcss. In v0.4 there is no need to use config file, anything can be configured in css. It would be useful if we can exclude config file when building tailwindcss.

### LINKED ISSUE #426: Cargo leptos build loads wrong version of the tailwingcss binary
# Cargo leptos build loads wrong tailwindcss (v4.0.0) binary

HOST: arm64
Docker image: **rust:alpine3.20** - use musl libc

Cargo leptos build loads a wrong tailwindcss binary when building leptos application in the ``rust:alpine3.20`` Docker container (->musl)

Dockerfile:
```
FROM rust:alpine3.20 AS builder 
...
RUN cargo install --locked cargo-leptos
...
# Set compilation environment 
ENV LEPTOS_TAILWIND_VERSION=${LEPTOS_TAILWIND_VERSION}
#ENV LEPTOS_SASS_VERSION=${LEPTOS_SASS_VERSION}
ENV LEPTOS_WASM_OPT_VERSION=${LEPTOS_WASM_OPT_VERSION}
ENV WASM_BIND_GEN=${WASM_BIND_GEN}
...
RUN cargo leptos build --release -vv
..
```

Compilation fails when it start process css files. This example has been run many times and the tailwindcss-linux-arm64 binary has been loaded in the first run.

```
217.4 Tailwind config: TailwindConfig { input_file: "style/tailwind.css", config_file: "tailwind.config.js", tmp_file: "/app/target/tmp/tailwind.css" } 
217.4 Command [tailwindcss] is_force_pin_version: true - Ok("v4.0.0") 
217.4 Command [tailwindcss] checking for the latest available version 
217.4 Style not configured 217.6 Command [tailwindcss] requested version v4.0.0 is already same or newer than available version v4.0.0 217.6 Command cache dir: /root/.cache/cargo-leptos 217.6 Command installing tailwindcss-v4.0.0 ... 
217.6 Install downloading tailwindcss https://github.com/tailwindlabs/tailwindcss/releases/download/v4.0.0/tailwindcss-linux-arm64 
218.9 Install decompressing tailwindcss /root/.cache/cargo-leptos/tailwindcss-v4.0.0/tailwindcss-v4.0.0 218.9 Command tailwindcss-v4.0.0 installed. 
218.9 Command using tailwindcss v4.0.0 /root/.cache/cargo-leptos/tailwindcss-v4.0.0/tailwindcss-v4.0.0/tailwindcss-linux-arm64 
218.9 Error: No such file or directory (os error 2)

```
In alpine environment builder should load **tailwindcss-linux-arm64-musl** binary instead of tailwindcss-linux-arm64 (this is linked with a glibc).

## Workaround
Install tailwindcss-linux-arm64-musl manually before the cargo leptos build command in a dockerfile.

Dockerfile - image: rust:alpine3.20 image

```
FROM rust:alpine3.20 AS builder 
...
...
# Leptos build loads wrong binary file: tailwindcss-linux-arm64 (linked with glibc) 
# Install the correct tailwindcss v4.0.0 binary manually 
RUN rm -rf /root/.cache/cargo-leptos/tailwindcss-v4.0.0 && \
    curl -LO https://github.com/tailwindlabs/tailwindcss/releases/download/v4.0.0/tailwindcss-linux-arm64-musl && \
    chmod +x tailwindcss-linux-arm64-musl && \
    mv tailwindcss-linux-arm64-musl /usr/local/bin/tailwindcss

RUN cargo leptos build --release -vv
...

```

SleeplessOne1917:
Closes #426 and closes #417.

Judging by the linked issues, I'm not the only one interested in using Tailwind v4 with Leptos. I added code to support use on linux musl, though I'm not sure how to test that it's correct. I also haven't figured out how to prevent the generation of a JS config file (which isn't needed in Tailwind v4) when using v4. Because of all of this, I'm opening this PR as a draft.


src/ext/exe.rs
```diff
@@ -621,12 +624,7 @@ trait Command {
     fn github_owner(&self) -> &str;
     fn github_repo(&self) -> &str;
     fn download_url(&self, target_os: &str, target_arch: &str, version: &str) -> Result<String>;
-    fn executable_name(
-        &self,
-        target_os: &str,
-        target_arch: &str,
-        version: Option<&str>,
-    ) -> Result<String>;
+    fn executable_name(&self, target_os: &str, target_arch: &str, version: &str) -> Result<String>;
```
SleeplessOne1917, line 627: Made `version` a `&str` instead of an `Option<&str>` since the only place the function is called, the version is guarunteed to be known. It also doesn't seem like this function will be called in a different place.

src/ext/exe.rs
```diff
@@ -331,6 +331,8 @@ impl Command for CommandTailwind {
 
     /// Tool binary download url for the given OS and platform arch
     fn download_url(&self, target_os: &str, target_arch: &str, version: &str) -> Result<String> {
+        let use_musl = is_linux_musl_env() && version.starts_with("v4");
```
SleeplessOne1917, line 334: Not sure if I should put in the effort to make the Tailwind version check more robust. I doubt they'll be coming out with a new major version any time soon.

Eveeifyeve:
>  I also haven't figured out how to prevent the generation of a JS config file (which isn't needed in Tailwind v4) when using v4.

Here: https://github.com/leptos-rs/cargo-leptos/blob/b2136f1e65e1b4a7f64112c1f718b8e8bfcbd2a4/src/config/tailwind.rs#L14-L28

> Because of all of this, I'm opening this PR as a draft.

Please mark it as a draft pr and not ready for review just to communicate that this is still work in progress and not ready to review.

benwis:
How are you planning on telling Tailwind to scan .rs files without a tailwind config file?

SleeplessOne1917:
@Eveeifyeve 
> Please mark it as a draft pr and not ready for review just to communicate that this is still work in progress and not ready to review.

My apologies. I accidentally opened it as ready for review and was unable to find out how to revert it to a draft PR after the fact.

@benwis 
> How are you planning on telling Tailwind to scan .rs files without a tailwind config file?

[According to the Tailwind CSS docs, almost every type of file gets scanned](https://tailwindcss.com/docs/detecting-classes-in-source-files#which-files-are-scanned):

> Tailwind will scan every file in your project for class names, except in the following cases:
> - Files that are in your `.gitignore` file
> - Binary files like images, videos, or zip files
> - CSS files
> - Common package manager lock files
>
> If you need to scan any files that Tailwind is ignoring by default, you can [explicitly register](https://tailwindcss.com/docs/detecting-classes-in-source-files#explicitly-registering-sources) those sources.

If the rust files still get ignored, we can still add sources explicity wiith `@source` in the CSS file.

benwis:
> @Eveeifyeve
> 
> > Please mark it as a draft pr and not ready for review just to communicate that this is still work in progress and not ready to review.
> 
> My apologies. I accidentally opened it as ready for review and was unable to find out how to revert it to a draft PR after the fact.
> 
> @benwis
> 
> > How are you planning on telling Tailwind to scan .rs files without a tailwind config file?
> 
> [According to the Tailwind CSS docs, almost every type of file gets scanned](https://tailwindcss.com/docs/detecting-classes-in-source-files#which-files-are-scanned):
> 
> > Tailwind will scan every file in your project for class names, except in the following cases:
> > 
> > * Files that are in your `.gitignore` file
> > * Binary files like images, videos, or zip files
> > * CSS files
> > * Common package manager lock files
> > 
> > If you need to scan any files that Tailwind is ignoring by default, you can [explicitly register](https://tailwindcss.com/docs/detecting-classes-in-source-files#explicitly-registering-sources) those sources.
> 
> If the rust files still get ignored, we can still add sources explicity wiith `@source` in the CSS file.

Good to know, I have not read their new docs

SleeplessOne1917:
From what I can tell, the JS config file is autogenerated in [`src/compile/tailwind.rs`](https://github.com/leptos-rs/cargo-leptos/blob/main/src/compile/tailwind.rs#L17-L19) as opposed to `src/config/tailwind.rs` as linked in a previous comment. The code in `src/config/tailwind.rs` does generate a `TailwindConfig` instance, but the only info `TailwindConfig::new` has to generate that is a `&ProjectConfig`, which doesn't contain any info directly about which tailwind version is being used. The current version is used in [`src/ext/exe.rs`](https://github.com/leptos-rs/cargo-leptos/blob/main/src/ext/exe.rs#L801-L803), although I couldn't find a clean way of making that accessible to code in other modules.

I could use the environment variable to tell what the version is, although that will break if the env var isn't specified and the user is using the default version. Alternatively, since `&ProjectConfig` being passed to `TailwindConfig::new` has the path of the input tailwindcss file, I could make read the first line of the file to see if it equals `@import "tailwindcss";`. Neither of these solutions are ideal, though.

SleeplessOne1917:
I decided to handle it with the `VersionConfig` enum in `src/config/version.rs`. Let me know if this is an acceptable way of managing it.

SleeplessOne1917:
Just remembered that preventing the JS config from generating when using v4 was the reason I marked this as draft. Now that that is handled, I'm marking this as ready for review again.

src/ext/exe.rs
benwis: I might consider making it error if the version is greater than 4 and provide a useful error message

benwis:
I think this is ready to merge once we make it error on tailwind v5. Nice work @SleeplessOne1917

benwis (approved):
LGTM!

benwis:
Thanks @SleeplessOne1917 I know the Tailwind using folks appreciate ya

PR #429, opened 2025-02-03
Use libraries instead of locally installed binaries for Cargo Generate and Wasm Opt

SleeplessOne1917:
While working on the Tailwind v4 PR, I noticed that there were some things that are handled with separate binaries that have libraries that do the same thing. I took the liberty of switching to using those libraries. This might also make it easier to handle [the version issue I an into with the Tailwind v4 PR](https://github.com/leptos-rs/cargo-leptos/pull/428#issuecomment-2629199292).

Even though this PR says it has a lot of additions, almost all of them are from the lockfile. As far as actual source code is concerned, this allowed me to rip out a lot of now unnecessary code.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ categories = ["development-tools", "wasm", "web-programming"]
 keywords = ["leptos"]
 version = "0.2.27"
 edition = "2021"
-rust-version = "1.71"
+rust-version = "1.82.0"
 authors = ["Henrik Akesson", "Greg Johnston", "Ben Wishovich"]
 
 [package.metadata.wix]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -89,6 +89,7 @@ swc = "10.0"
 swc_common = "5.0"
 shlex = "1.3.0"
 cargo-generate = "0.22"
+wasm-opt = "0.116.1"
 
 [dev-dependencies]
 insta = { version = "1.40.0", features = ["yaml"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -88,6 +88,7 @@ base64ct = { version = "1.6.0", features = ["alloc"] }
 swc = "10.0"
 swc_common = "5.0"
 shlex = "1.3.0"
+cargo-generate = "0.22"
 
 [dev-dependencies]
 insta = { version = "1.40.0", features = ["yaml"] }
--- a/src/compile/front.rs
+++ b/src/compile/front.rs
@@ -15,7 +15,7 @@ use swc::JsMinifyExtras;
 use swc::{config::JsMinifyOptions, try_with_handler, BoolOrDataConfig};
 use swc_common::{FileName, SourceMap, GLOBALS};
 use tokio::process::Child;
-use tokio::{process::Command, sync::broadcast, task::JoinHandle};
+use tokio::{process::Command, task::JoinHandle};
 use wasm_bindgen_cli_support::Bindgen;
 use wasm_opt::OptimizationOptions;
 
@@ -105,7 +105,6 @@ pub fn build_cargo_front_cmd(
 
 async fn bindgen(proj: &Project) -> Result<Outcome<Product>> {
     let wasm_file = &proj.lib.wasm_file;
-    let interrupt = Interrupt::subscribe_any();
 
     log::info!("Front generating JS/WASM with wasm-bindgen");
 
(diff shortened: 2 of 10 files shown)


Cargo.toml
```diff
@@ -31,7 +31,7 @@ categories = ["development-tools", "wasm", "web-programming"]
 keywords = ["leptos"]
 version = "0.2.27"
 edition = "2021"
-rust-version = "1.71"
+rust-version = "1.82.0"
```
SleeplessOne1917, line 34: Clippy was giving warnings that the earliest rust version to support [`is_none_or` (which is used in this project)](https://github.com/leptos-rs/cargo-leptos/blob/main/src/config/bin_package.rs#L90) is 1.82.0.

src/compile/front.rs
```diff
@@ -193,18 +187,10 @@ async fn bindgen(proj: &Project) -> Result<Outcome<Product>> {
...
-
-    let args = [file.as_str(), "-Oz", "-o", file.as_str()];
-    let process = Command::new(wasm_opt)
-        .args(args)
-        .spawn()
-        .context("Could not spawn command")?;
-    wait_interruptible("wasm-opt", process, interrupt).await
+fn optimize(file: &Utf8Path) -> Result<()> {
+    OptimizationOptions::new_optimize_for_size_aggressively()
+        .run(file, file)
```
SleeplessOne1917, line 192: Not sure how to make this interruptable since `run` is synchronous. Unsure how much it's needed since it's not a separate process anymore.

src/compile/tailwind.rs
```diff
@@ -18,7 +18,7 @@ pub async fn compile_tailwind(proj: &Project, tw_conf: &TailwindConfig) -> Resul
         create_default_tailwind_config(tw_conf).await?;
     }
 
-    let (line, process) = tailwind_process(&proj, "tailwindcss", tw_conf).await?;
+    let (line, process) = tailwind_process(proj, "tailwindcss", tw_conf).await?;
```
SleeplessOne1917, line 21: This is to fix a Clippy warning.

src/ext/exe.rs
```diff
@@ -76,7 +75,7 @@ pub struct ExeCache<'a> {
     meta: &'a ExeMeta,
 }
 
-impl<'a> ExeCache<'a> {
+impl ExeCache<'_> {
```
SleeplessOne1917, line 78: Fixes a Clippy warning.

src/ext/exe.rs
```diff
@@ -269,47 +253,48 @@ impl Exe {
...
 /// We generally want to keep the suffix intact,
 /// as it carries classifiers, etc, but strip non-ascii
 /// digits from the prefix.
 #[inline]
-fn sanitize_version_prefix(ver_string: &str) -> String {
-    ver_string
-        .chars()
-        .skip_while(|c| !c.is_ascii_digit() || *c == '_')
-        .collect::<String>()
+fn sanitize_version_prefix<'a>(ver_string: &'a str) -> Result<&'a str> {
```
SleeplessOne1917, line 260: Since `wasm-opt` is now used as a library instead of a binary downloaded from Github, I figured the check for "version_" is unnecessary.

src/lib.rs
```diff
@@ -72,11 +72,11 @@ pub async fn run(args: Cli) -> Result<()> {
     let _monitor = Interrupt::run_ctrl_c_monitor();
     use Commands::{Build, EndToEnd, New, Serve, Test, Watch};
     match args.command {
-        New(_) => panic!(),
         Build(_) => command::build_all(&config).await,
         Serve(_) => command::serve(&config.current_project()?).await,
         Test(_) => command::test_all(&config).await,
         EndToEnd(_) => command::end2end_all(&config).await,
         Watch(_) => command::watch(&config.current_project()?).await,
+        New(_) => unreachable!(r#""new" command should have already been run"#),
```
SleeplessOne1917, line 80: I figure `unreachable!` is closer to what is meant semantically here. The docs for that macro say that it's useful whenver the compiler can't determine if some code is unreachable; I'm unsure if that means it will aid the compiler in optimization, but it seems more appropriate regardless.

SleeplessOne1917:
I made a [PR for `wasm-opt`](https://github.com/brson/wasm-opt-rs/pull/167) to update it to use a recent version of binaryen.

benwis:
@SleeplessOne1917 I like it! Anything to make our lives a little easier. Ironically I'm pretty sure I want to not use wasm-bindgen's library because it causes version mismatch error. I don't think that'll happen in this case, but something to think about.

Got a few conflicts too

SleeplessOne1917:
> I'm pretty sure I want to not use wasm-bindgen's library because it causes version mismatch error.

As far as I can tell, [`wasm-opt`](https://github.com/brson/wasm-opt-rs) isn't related to `wasm-bindgen`. At the very least, the 2 libraries belong to different GitHub users.

benwis:
> > I'm pretty sure I want to not use wasm-bindgen's library because it causes version mismatch error.
> 
> As far as I can tell, [`wasm-opt`](https://github.com/brson/wasm-opt-rs) isn't related to `wasm-bindgen`. At the very least, the 2 libraries belong to different GitHub users.

Oh I know they're not related, its just interesting that we're going different directions with tooling. A few more conflicts here

benwis (approved):
I like it, all good changes

PR #431, opened 2025-02-03
Remove dep on ansi_term

SleeplessOne1917:
Running `cargo-audit` on this repo revealed that it depends directly on an unmaintained dependency: `ansi_term`. This PR removes that dependency. Fortunately, it just so happens that `clap`—which `cargo-leptos` already depends on—supports everything `cargo-leptos` was using `ansi_term` to do.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -104,15 +104,6 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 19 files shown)


benwis (approved):
LGTM

Issue #430, opened 2025-02-03
`cargo watch` No longer auto reloading since 0.2.25


luxalpa:
Since version 0.2.25 `cargo-leptos watch` does not automatically recompile the code and restart the server on Windows when it detects changes. It seems like it doesn't detect changes at all. It still works in 0.2.24. The problem still exists in 0.2.27.

After some testing, I can confirm it is due to this commit: https://github.com/leptos-rs/cargo-leptos/commit/b900521f4e7c8e51773e9b640412a59d40406cb8

@sabify

sabify:
My best guess is that line https://github.com/leptos-rs/cargo-leptos/blob/b2136f1e65e1b4a7f64112c1f718b8e8bfcbd2a4/src/service/notify.rs#L98 causes the problem and the event will be ignored sent by Windows's `ReadDirectoryChangesW` API. I don't have Windows to test it out, but could you please add the following code before `return` and show me the logs here?

```rust
log::error!("Notify ignored the event {:?}", event);
```

luxalpa:
```
listening on http://127.0.0.1:3000
      Notify ignored the event Event { kind: Modify(Any), paths: ["C:\\projects\\leptos-test\\src\\app.rs~"], attr:tracker: None, attr:flag: None, attr:info: None, attr:source: None }
      Notify ignored the event Event { kind: Modify(Any), paths: ["C:\\projects\\leptos-test\\src\\app.rs~"], attr:tracker: None, attr:flag: None, attr:info: None, attr:source: None }
      Notify ignored the event Event { kind: Modify(Any), paths: ["C:\\projects\\leptos-test\\src\\app.rs"], attr:tracker: None, attr:flag: None, attr:info: None, attr:source: None }
      Notify ignored the event Event { kind: Modify(Any), paths: ["C:\\projects\\leptos-test\\src\\app.rs"], attr:tracker: None, attr:flag: None, attr:info: None, attr:source: None }
      Notify ignored the event Event { kind: Modify(Any), paths: ["C:\\projects\\leptos-test\\src\\app.rs"], attr:tracker: None, attr:flag: None, attr:info: None, attr:source: None }
```

sabify:
Thanks for sharing, my guess was right. I will create a patch for this.

sabify:
Could you please check this PR (https://github.com/leptos-rs/cargo-leptos/pull/432) out and see if it is working?

luxalpa:
Yes, it worked. Thanks!

PR #432, opened 2025-02-03
fix: catch `any` event in case the OS isn't specific about the event.

sabify:
Windows isn't specific about the notify event.
Closes: #430


gbj:
Thank you!

PR #433, opened 2025-02-04
refactor: Moved logging init into main.rs

ActuallyHappening:
This allows me to link against the library without `panic`ing on twice initialised logging (I'm using tracing)


benwis (approved):
@ActuallyHappening Thanks for the PR! I'm glad to make this a little easier for ya.

PR #437, opened 2025-02-10
Upgrade to `tracing` and `color_eyre`

ActuallyHappening:
Build upon https://github.com/leptos-rs/cargo-leptos/pull/433
Uses tracing for collecting logs, in the future somebody can upgrade the outputting infrastructure (for non-library linked use cases) to use `tracing-subscriber`, at the moment it works exactly as before.
Upgraded to `color_eyre` from `anyhow`. This means errors are reported in beautiful colour by default!
Other than the API exposed by the library, there should be no noticable difference.
See also my similar effort in `pueue`: https://github.com/Nukesor/pueue/pull/602


ActuallyHappening:
I didn't need to `cargo update`, you can drop that commit and only upgrade `swc` and `swc_common` deps if desired

benwis:
Can you run `cargo build` and upload the requisite files?

ActuallyHappening:
> Can you rebase this onto latest main?

I'm not very good with git, is my latest forced push enough?

benwis:
Thanks for the PR, I appreciate the changes. Out of curiosity, what are you consuming cargo-leptos as a lib in?

ActuallyHappening:
> Out of curiosity, what are you consuming cargo-leptos as a lib in?
I'm using cargo subcommands as my developer tooling, `cargo xtask serve` for example, which means being able to pull in cargo-leptos as a rust library avoids a level of indirection using child processes that makes me happy :)
Plus, since the `xtask` crate and my actual app code are in the same workspace, they are pinned to the same `wasm-bindgen` library version, so I'm guarenteed to not have a version mismatch, which is one symptom of correct developer tooling like this

sabify:
@benwis This PR unexpectedly reintroduced patch.rs! The file should have left removed.

benwis:
Whoops, removed it again

ActuallyHappening:
Sorry! That would have been my lack of git knowledge :)

PR #408, opened 2024-12-19
fix: do not panic on interrupt channel for watch

sabify:
Closes: https://github.com/leptos-rs/cargo-leptos/issues/385

PR #439, opened 2025-02-14
fix(watch): add debounce to notify

sabify:
This PR does:
- aggregates the change events from notify (every 100 ms) and then tries to send appropriate signals to build loops.
- prevents overloading the interrupt-based cycle, especially while modifying multiple files.
- makes the event loop more efficient.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -588,7 +588,7 @@ dependencies = [
  "lightningcss",
  "log",
  "md-5",
- "notify",
+ "notify-debouncer-full",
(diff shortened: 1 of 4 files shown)


sabify:
This may possibly solve the https://github.com/leptos-rs/cargo-leptos/pull/408#issuecomment-2626678795 issue.

benwis:
Let's gives this a shot, I can see the utility here. Thanks for the PR

benwis (approved):

PR #449, opened 2025-02-23
Fix (some) cargo dist releases

jheuel:
### Problem
All `aarch64` builds and the `musl` build are [currently failing](https://github.com/leptos-rs/cargo-leptos/actions/runs/13457858276) in our `cargo-dist` setup.

### Changes in this PR
First, I saw that `cargo-dist` is `dist` now and moved the configuration out of `Cargo.toml` into `dist-workflow.toml`, so I followed their recommendation.

Looking at the build logs, we see that the `aarch64` builds are complaining about not finding the OpenSSL installation. So I added a vendored OpenSSL, which fixes the Darwin and Linux builds. 

The `aarch64` Windows build, however, fails while trying to compile `psm v0.1.25`, which [does not support Windows](https://github.com/rust-lang/stacker/issues/99#issuecomment-2404989304).

`x86_64-unknown-linux-musl` fails because it cannot find `musl-g++`. `musl` does not package a C++ compiler. I tried the [ugly fix of symlinking another g++ to musl-g++](https://github.com/rust-lang/cargo/issues/3359#issuecomment-325614445) but then it fails later because it cannot find some fortified functions while linking. I guess one has to manually compile a C++ compiler targetting `musl`.

### Conclusion
I would suggest disabling the `aarch64-pc-windows-msvc` and `x86_64-unknown-linux-musl` builds for now in cargo dist.

### DIFF
--- a/.github/build-setup.yml
+++ b/.github/build-setup.yml
@@ -1,0 +1,10 @@
+- name: Install MUSL tools and configure environment
+  env:
+    TARGETS: ${{ join(matrix.targets, '-') }}
+  if: contains(env.TARGETS, 'x86_64-unknown-linux-musl')
+  run: |
+    sudo apt-get update
+    sudo apt-get install -y musl-tools musl-dev build-essential pkg-config libssl-dev
+    if ! command -v musl-g++ > /dev/null; then
+      sudo ln -s $(which musl-gcc) /usr/local/bin/musl-g++
+    fi
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -125,6 +125,13 @@ jobs:
             curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
             echo "$HOME/.cargo/bin" >> $GITHUB_PATH
           fi
+      - name: "Symlink g++ compiler for musl targets"
+        if: "contains(env.TARGETS, 'x86_64-unknown-linux-musl')"
+        run: |
+          sudo ln -s /usr/bin/g++ /usr/bin/musl-g++
+          ls -l /usr/bin/musl-g++
+        env:
+          "TARGETS": "${{ join(matrix.targets, '-') }}"
       - name: Install dist
         run: ${{ matrix.install_dist.run }}
       # Get the dist-manifest
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -125,11 +125,14 @@ jobs:
             curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
             echo "$HOME/.cargo/bin" >> $GITHUB_PATH
           fi
-      - name: "Symlink g++ compiler for musl targets"
+      - name: "Install MUSL tools and configure environment"
         if: "contains(env.TARGETS, 'x86_64-unknown-linux-musl')"
         run: |
-          sudo ln -s /usr/bin/g++ /usr/bin/musl-g++
-          ls -l /usr/bin/musl-g++
+          sudo apt-get update
+          sudo apt-get install -y musl-tools musl-dev build-essential pkg-config libssl-dev
+          if ! command -v musl-g++ > /dev/null; then
+            sudo ln -s $(which musl-gcc) /usr/local/bin/musl-g++
+          fi
         env:
           "TARGETS": "${{ join(matrix.targets, '-') }}"
       - name: Install dist
--- a/dist-workspace.toml
+++ b/dist-workspace.toml
@@ -23,3 +23,5 @@ hosting = ["axodotdev", "github"]
 install-path = "CARGO_HOME"
 # Whether to install an updater program
 install-updater = false
+# Add custom build setup steps
+github-build-setup = "../build-setup.yml"
--- a/dist-workspace.toml
+++ b/dist-workspace.toml
@@ -8,7 +8,7 @@ cargo-dist-version = "0.28.0"
 # CI backends to support
 ci = "github"
 # Target platforms to build apps for (Rust target-triple syntax)
-targets = ["aarch64-apple-darwin", "aarch64-unknown-linux-gnu", "aarch64-pc-windows-msvc", "x86_64-apple-darwin", "x86_64-unknown-linux-gnu", "x86_64-unknown-linux-musl", "x86_64-pc-windows-msvc"]
+targets = ["aarch64-apple-darwin", "aarch64-unknown-linux-gnu", "x86_64-apple-darwin", "x86_64-unknown-linux-gnu", "x86_64-unknown-linux-musl", "x86_64-pc-windows-msvc"]
 # The installers to generate for each app
 installers = ["shell", "powershell"]
 # Which actions to run on pull requests
@@ -18,7 +18,7 @@ windows-archive = ".tar.gz"
 # The archive format to use for non-windows builds (defaults .tar.xz)
 unix-archive = ".tar.gz"
 # Where to host releases
-hosting = ["axodotdev", "github"]
+hosting = "github"
 # Path that installers should place binaries in
 install-path = "CARGO_HOME"
 # Whether to install an updater program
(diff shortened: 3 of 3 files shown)


Cargo.toml
```diff
@@ -92,6 +68,7 @@ cargo-generate = "0.22"
 wasm-opt = "0.116.1"
 ignore = "0.4.23"
 walkdir = "2.5"
+openssl = { version = "0.10.71", features = ["vendored"] }
```
s0l0ist, line 71: instead of adding the dep here, I think the `cargo-generate` dep has a feature flag for vendored openssl and git2 which maybe a better fix?

s0l0ist: I've confirmed this fixes the issue for me on Alpine (musl) builds for `aarch64-unknown-linux-musl`.

I had to add a few more deps to my docker image:

- make
- perl

s0l0ist: I've added https://github.com/leptos-rs/cargo-leptos/pull/452 for reference

.github/build-setup.yml
```diff
@@ -0,0 +1,8 @@
+# musl does not include a g++ compiler
+- name: Symlink g++ compiler for musl targets
+  env:
+    TARGETS: ${{ join(matrix.targets, '-') }}
+  if: contains(env.TARGETS, 'x86_64-unknown-linux-musl')
+  run: |
+    sudo ln -s /usr/bin/g++ /usr/bin/musl-g++
+    ls -l /usr/bin/musl-g++
```
sabify, line 8: @jheuel Can you please use the following script instead?

```yml
      - name: Install MUSL tools and configure environment
        if: contains(matrix.target, 'musl')
        run: |
          sudo apt-get update
          sudo apt-get install -y musl-tools
          if ! command -v musl-g++ >/dev/null; then
            sudo ln -s $(which musl-gcc) /usr/local/bin/musl-g++
          fi
```

.github/workflows/release.yml
```diff
@@ -127,6 +127,13 @@ jobs:
             curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
             echo "$HOME/.cargo/bin" >> $GITHUB_PATH
           fi
+      - name: "Symlink g++ compiler for musl targets"
+        if: "contains(env.TARGETS, 'x86_64-unknown-linux-musl')"
+        run: |
+          sudo ln -s /usr/bin/g++ /usr/bin/musl-g++
+          ls -l /usr/bin/musl-g++
+        env:
+          "TARGETS": "${{ join(matrix.targets, '-') }}"
```
sabify, line 136: @jheuel This file doesn't need to be modified.

dist-workspace.toml
```diff
@@ -0,0 +1,27 @@
...
+# The archive format to use for non-windows builds (defaults .tar.xz)
+unix-archive = ".tar.gz"
+# Where to host releases
+hosting = ["axodotdev", "github"]
+# Path that installers should place binaries in
+install-path = "CARGO_HOME"
+# Whether to install an updater program
+install-updater = false
+# Add custom build setup steps
+github-build-setup = "../build-setup.yml"
```
sabify, line 27: @jheuel Please rebase and keep this line and the `targets` one.

sabify:
Hi @jheuel, Thanks for the PR. Can you please rebase and follow the comments I've mentioned?

.github/build-setup.yml
jheuel: I'll change it. However, cargo dist already installs `musl-tools`, so I would drop the apt-get lines.

.github/workflows/release.yml
jheuel: This modification was done by cargo dist, the whole file is generated by it

.github/build-setup.yml
sabify: Please keep it. It would be predictable.

.github/workflows/release.yml
sabify: Please regenerate it.

jheuel:
Will do later today.

Cargo.toml
sabify: We already added `libssl-dev` as part of build process. Please remove this line too.

jheuel:
Seems like the new trigger in `build-setup.yml` does not work.

sabify:
Please remove `if: contains(matrix.target, 'musl')` from the script, regenerate and test it again.

jheuel:
I tried with the old setup with the environment variable to only fire on the musl build. Full log: https://github.com/jheuel/cargo-leptos/actions/runs/13993610645/job/39183148524

Now fails with:
```
  cargo:warning=In file included from /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cxx-1.0.148/src/cxx.cc:1:
  cargo:warning=/home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cxx-1.0.148/src/../include/cxx.h:2:10: fatal error: algorithm: No such file or directory
  cargo:warning=    2 | #include <algorithm>
  cargo:warning=      |          ^~~~~~~~~~~
  cargo:warning=compilation terminated.

  --- stderr


  error occurred in cc-rs: command did not execute successfully (status code exit status: 1): LC_ALL="C" "musl-g++" "-O3" "-ffunction-sections" "-fdata-sections" "-fPIC" "-m64" "-static" "-std=c++11" "-Wall" "-Wextra" "-o" "/home/runner/work/cargo-leptos/cargo-leptos/target/x86_64-unknown-linux-musl/dist/build/cxx-c31abd394c24a950/out/c16f17691ff6f04b-cxx.o" "-c" "/home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cxx-1.0.148/src/cxx.cc"
```

jheuel:
After the openssl vendoring and the new cargo dist settings setup were already merged into main, this PR now only contains our current best attempt to build for linux musl.
It still fails the release workflow.

benwis:
> After the openssl vendoring and the new cargo dist settings setup were already merged into main, this PR now only contains our current best attempt to build for linux musl. It still fails the release workflow.

It appears the plan CI task below was successful, and if  you rebased on main that one failed windows test should pass. So I'm not sure what you mean

sabify:
@jheuel Here is the script:

```yaml
- name: "Install MUSL tools and configure environment"
  if: "contains(join(matrix.targets, ' '), 'musl')"
  run: |
    sudo apt-get update
    sudo apt-get install -y musl musl-dev musl-tools clang
    sudo ln -s /usr/bin/g++ /usr/bin/musl-g++
```

Please replace it and regenerate, it will succeed.

sabify:
Wait please, got an error in linking!

https://github.com/sabify/cargo-leptos/actions/runs/14001786575/job/39209536568

One more is missing...

benwis:
It remains odd to me that this used to work perfectly fine with no effort whatsoever, and now we've got to jump through hoops

jheuel:
> It appears the plan CI task below was successful, and if you rebased on main that one failed windows test should pass. So I'm not sure what you mean

Are you talking about the CI tests task? They do not touch anything this PR changes.

> Wait please, got an error in linking!

Yes, with those changes we have about the same again what I had before and we get the [same error](https://github.com/jheuel/cargo-leptos/actions/runs/13629135760/job/38092853073). Maybe we can rid of the dependency that requires the things that are missing?

jheuel:
> It remains odd to me that this used to work perfectly fine with no effort whatsoever, and now we've got to jump through hoops

`wasm-opt` requires a C++ compiler and standard library to link against, which was added since the last time cargo-leptos build correctly on linux musl (v0.2.26, [Cargo.toml diff](https://github.com/leptos-rs/cargo-leptos/compare/v0.2.26..main#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542))

sabify:
Please checkout #471.

jheuel:
Yes, another base image is smarter. Closing this one

PR #452, opened 2025-03-02
Fix `aarch64-linux-unknown-musl` by using vendored-openssl in cargo-generate

s0l0ist:
### Issue
Compiling cargo-leptos is broken in version v0.2.29 due to OpenSSL-related dependencies for the `aarch64-linux-unknown-musl` target. I've successfully compiled and verified functionality after applying this fix using the following docker image:

```docker
FROM public.ecr.aws/docker/library/alpine:latest

RUN apk update && \
    apk add --no-cache \
    bash \
    git \
    binaryen \
    gcc \
    g++ \
    perl \
    make \
    npm \
    curl

# Install Rust + cargo-leptos
RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
ENV PATH="/root/.cargo/bin:${PATH}"
RUN rustup target add wasm32-unknown-unknown

# This PR's changes in my fork:
# RUN cargo install --git https://github.com/s0l0ist/cargo-leptos cargo-leptos --locked
...
```

### Impact
The feature flag relies on the version pinned by carg-generate's dependencies and may be slightly older than the `openssl` crate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -550,6 +550,7 @@ dependencies = [
  "liquid-lib",
  "log",
  "names",
+ "openssl",
  "paste",
  "path-absolutize",
  "regex",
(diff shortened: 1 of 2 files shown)


paul-hansen (approved):
This would also fix CI builds taking 25 min extra due to binstall/quickinstall not being able to build cargo-leptos since 0.2.29

I actually made a PR with the exact same changes before I noticed this one: https://github.com/leptos-rs/cargo-leptos/pull/461/files

Issue #436, opened 2025-02-08
Add option to skip one or more of the phases during build.


twitchax:
I am happy to send a PR for this, but want to make sure it would be viable.

My use case is that I am building my entire leptos + axum stack into a WASI binary that then executes in `wasmtime` (e.g., https://github.com/twitchax/leptos-axum-wasi).  `cargo leptos build` does _most_ of the heavy lifting, but I end up duplicating the server-side build using the `wasm32-wasip2` target.

So, either `cargo leptos build` needs to support a target flag, or it would be nice to "save" on some of the work by skipping the server-side build.  As of this moment, to prepare my container, I run `cargo leptos build` which generates my frontend + the `site` directory, then builds an x86 backend binary.  Then, I build the server binary again with the wasm target (i.e., `cargo build --lib --release --target wasm32-wasip2 --no-default-features --features ssr`).

So, it would be cool if I could `--skip-backend` (or `--skip-frontend`, or `--skip-site-root-copy`, etc.).

If this seems reasonable, I am happy to do the minimal PR for `--skip-backend`.

benwis:
I would definitely be interested in a --lib-only or --bin--only flag, so if you'd like to PR that it'd be great

benwis:
Closing as duplicate of #315

PR #454, opened 2025-03-05
Feat: --bin-only / --lib-only flags

broose-goose:
In response to https://github.com/leptos-rs/cargo-leptos/issues/436.

Naive approach to just get it working. I think it actually is, but doesn't quite work with [my use case](https://github.com/broose-goose/leptos-multi-lib-workspace)... If I get that working, I'll polish this up, document the approach, and merge this


benwis:
@broose-goose Don't give up, it might not be for your use case but does it actually build either the lib or the bin?

benwis:
I'm closing this as stale, feel free to reopen and fix the conflicts if you'd still like this

PR #455, opened 2025-03-10
Update version of example to 0.7.x

SleeplessOne1917:
Since 0.7.x has been stable for awhile, I think it makes sense to make the example for `cargo-leptos` use it.


benwis:
Sure, seems like a plan. I've always debated whether to keep this one since it usually tracks way behind the proper ones

PR #456, opened 2025-03-10
Add documentation for some env vars/config options

SleeplessOne1917:
I brought this up in the Discord channel for `cargo-leptos`.

### DIFF
--- a/src/config/project.rs
+++ b/src/config/project.rs
@@ -167,16 +167,19 @@ impl Project {
             vec.push(("LEPTOS_HASH_FILE_NAME", self.hash_file.rel.to_string()));
         }
         if self.watch {
-            vec.push(("LEPTOS_WATCH", true.to_string()))
+            vec.push(("LEPTOS_WATCH", self.watch.to_string()))
         }
         if let Some(prefix) = self.server_fn_prefix.as_ref() {
             vec.push(("SERVER_FN_PREFIX", prefix.clone()));
         }
         if self.disable_server_fn_hash {
-            vec.push(("DISABLE_SERVER_FN_HASH", true.to_string()));
+            vec.push((
+                "DISABLE_SERVER_FN_HASH",
+                self.disable_server_fn_hash.to_string(),
+            ));
         }
         if self.server_fn_mod_path {
-            vec.push(("SERVER_FN_MOD_PATH", true.to_string()));
+            vec.push(("SERVER_FN_MOD_PATH", self.server_fn_mod_path.to_string()));
         }
         vec
     }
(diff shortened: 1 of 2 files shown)


src/config/project.rs
```diff
@@ -167,16 +167,19 @@ impl Project {
...
         if self.disable_server_fn_hash {
-            vec.push(("DISABLE_SERVER_FN_HASH", true.to_string()));
+            vec.push((
+                "DISABLE_SERVER_FN_HASH",
+                self.disable_server_fn_hash.to_string(),
+            ));
         }
         if self.server_fn_mod_path {
-            vec.push(("SERVER_FN_MOD_PATH", true.to_string()));
+            vec.push(("SERVER_FN_MOD_PATH", self.server_fn_mod_path.to_string()));
```
SleeplessOne1917, line 182: I changed these to not use literals since boolean env vars seem to handled as `NAME=value` for this project.

benwis:
Thanks for the PR, I think it's worth expanding on how the JS is minified, what does it, and whether it can be configured. I've certainly got questions about it I do not know the answer to. Love everything else

benwis:
> I updated the description of `js-minify`. There weren't any comments to copy/paste like with the others, so I listed everything I could deduce.
> 
> Also, I noticed this in `project.rs`:
> 
> ```rust
> js_minify: cli.release && cli.js_minify && config.js_minify,
> ```
> 
> Is it just me, or does this seem off? I would think it would be something like `cli.js_minify || (cli.release && config.js_minify`. The current logic means that even if a user enables JS minification in their config and is building a binary for production, they still need to remember to pass the CLI option to minify JS. My suggested change makes it so that JS is minified if enabled in the config and building for production, but users can still use the CLI option to overwrite it when making a debug build.

Yeah that is definitely not correct

src/config/project.rs
```diff
@@ -130,7 +130,7 @@ impl Project {
                 watch_additional_files,
                 hash_file,
                 hash_files: config.hash_files,
-                js_minify: cli.release && cli.js_minify && config.js_minify,
+                js_minify: cli.release && (cli.js_minify || config.js_minify),
```
SleeplessOne1917, line 133: Ended up doing something slightly different than I originally proposed. [According to this comment in `cli.rs`](https://github.com/leptos-rs/cargo-leptos/blob/4b0c234bfd4efba05fbbfde2dbd625ab537c79e7/src/config/cli.rs#L59), the CLI option also only works for release builds. Now, it only ever works in release, but will work if either the CLI option or config option are specified.

PR #461, opened 2025-03-15
Enable vendored-openssl feature on cargo-generate

### LINKED ISSUE #460: binstall fails to find binaries for cargo-leptos 0.2.29
For cli tools like cargo-leptos, having a quick way to install it is very helpful for CI.

I've been using [cargo-binstall](https://github.com/cargo-bins/cargo-binstall) which installs crates from cargo-quickinstall's prebuilt binaries, or it pulls a binary from the github releases, finally falling back to compiling from source.

The latest github release is several versions behind, and 0.2.29 is failing to build on quickinstall. This means installing cargo-leptos falls back to building from source which makes running it in a Github actions really slow.

There is an existing issue with a response on quickinstall here: https://github.com/cargo-bins/cargo-quickinstall/issues/345
> https://github.com/cargo-bins/cargo-quickinstall/actions/runs/13536410650/job/37828818967
>
> There's a failure when compiling it, because it didn't find openssl
>
> The best fix is for cargo-leptos to turn on openssl/vendored, or not use openssl

paul-hansen:
Hopefully fixes quickinstall builds #345 and closes #460

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -550,6 +550,7 @@ dependencies = [
  "liquid-lib",
  "log",
  "names",
+ "openssl",
  "paste",
  "path-absolutize",
  "regex",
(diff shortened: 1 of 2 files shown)


jheuel:
Should work if you disable the `aarch64-pc-windows-msvc` and `x86_64-unknown-linux-musl` targets, which fail for other reasons. See https://github.com/leptos-rs/cargo-leptos/pull/452

paul-hansen:
Ah I missed that there were existing PRS for this from weeks ago. Will close this one.

jheuel:
I disabled the two target on my fork and let it [build v0.2.29](https://github.com/jheuel/cargo-leptos/releases/tag/v0.2.30) for the same reason.

However, then I found https://github.com/taiki-e/cache-cargo-install-action which installs from source once and then cashes the binary.

PR #468, opened 2025-03-21
Add ability to run custom post build scripts defined in build-scripts metadata

djkato:
usecase:
Bundling a Bevy app into my Leptos site caused 6mb time to first interaction delay, which I solved by creating a separate workspace library just for bevy, and importing it manually via JS. Problem was that every time I copied the WASM binary and ran `cargo leptos serve`, it would delete the library and I had to do it manually again.

With post build scripts I can:
-  `cargo leptos watch` the workspace
- add my library to `watch_additional_files`
- have leptos rebuild my project through my scripts, move the files to site/pkg
- know my js will be able to fetch the extra wasm files there every time.

Example how it looks like in Cargo.toml:
```toml
[[workspace.metadata.leptos]]
#...
build-scripts = [
  "cd project1/extras/extra-package-1 && wasm-pack build --release --target web --out-dir=../../../target/extra/package1",
  "cd project1/extras/extra-package-2 && wasm-pack build --release --target web --out-dir=../../../target/extra/package2",
  "mkdir -p ./target/site/project1/pkg/extra",
  "mv ./target/extra/package*/*.wasm ./target/site/project1/pkg/extra",
  "mv ./target/extra/package*/*.js ./target/site/project1/pkg/extra",
]
```
This is my first feature submission to a larger repo like this, please be gentle with the critique x)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14,24 +14,18 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
-version = "0.21.0"
+version = "0.25.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8a30b2e23b9e17a9f90641c7ab1549cd9b44f296d3ccbf309d2863cfe398a0cb"
+checksum = "1b5d307320b3181d6d7954e663bd7c774a838b8220fe0593c86d9fb09f498b4b"
 dependencies = [
- "gimli 0.28.1",
+ "gimli",
 ]
 
 [[package]]
-name = "adler"
-version = "1.0.2"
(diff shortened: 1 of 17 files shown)


sabify:
@djkato `build-scripts` needs to take a list of paths to scripts instead of the scripts’ content.

djkato:
@sabify it takes both. It just runs `sh -c {}`, so if a user wants to put it in a file they can define them as
"./my-script.sh" or "script.bat".
I'd rather allow both. Creating another whole file just to run for example "cp -r target/site site" seems annoying to me.

PR #474, opened 2025-03-29
fix: ensure the end2end command has exited on interrupt

### LINKED ISSUE #387: `end-to-end` tests leave ports open if tests fail
When I run `cargo leptos end-to-end`, if a test fails, when I attempt to run again, it fails to serve the application.

```
Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.20s
       Cargo finished cargo build --package=my-app --bin=my-app --no-default-features --features=ssr
     Serving at http://0.0.0.0:8000
thread 'main' panicked at code/rust/ui/my-app/src/main.rs:26:63:
called `Result::unwrap()` on an `Err` value: Os { code: 98, kind: AddrInUse, message: "Address already in use" }
```

ilyvion:
It does this by checking whether it's exited or not, and if not, kills it. This ensures that the command doesn't keep running in the background or cause any other odd shell behaviors once `cargo-leptos` returns.

Presumably resolves #387

### DIFF
--- a/src/command/end2end.rs
+++ b/src/command/end2end.rs
@@ -51,7 +51,20 @@ async fn try_run(cmd: &str, dir: &Utf8Path) -> Result<()> {
     let mut int = Interrupt::subscribe_any();
 
     tokio::select! {
-          _ = int.recv() => Ok(()),
+          _ = int.recv() => {
+            match process.try_wait() {
+                Ok(None) | Err(_) => {
+                    trace!("End2End child process still alive; attempting to kill...");
+                    match process.kill().await {
+                        Err(e) => warn!("Could not kill End2End child process: {}", e),
+                        Ok(_) => trace!("Successfully killed End2End child process"),
+                    }
+                }
+                Ok(Some(status)) => trace!("End2End child process already exited with {status}"),
+            }
+            
+            Ok(())
+        },
           result = process.wait() => {
             let status = result?;
             if !status.success() {


paul-hansen:
It seems like this isn't killing the Leptos server that `cargo-leptos end-to-end` is starting so the ports are still staying in use.

I tested this locally on my Axum project by installing cargo-leptos using 
```shell
cargo install cargo-leptos --git=https://github.com/ilyvion-contrib/cargo-leptos.git --branch=en2end-kill-cmd
```

ilyvion:
> I tested this locally on my Axum project by installing cargo-leptos using
> 
> ```shell
> cargo install cargo-leptos --git=https://github.com/ilyvion-contrib/cargo-leptos.git --branch=en2end-kill-cmd
> ```

Did you/could you try running it with `RUSTLOG="trace"` to see what the logging I added outputs?

paul-hansen:
Actually, I think the server only keeps running when the tests fail. If I make them all succeed it seems to kill it.

Note that cargo leptos doesn't seem to log it's logs based on RUST_LOG (I'd guess so you can use that for your app), you have to use `cargo leptos end-to-end -vvv` to enabling tracing logs. I still didn't see the new traces at all, though I'd guess the tests failing might be causing it to exit before it gets there?

ilyvion:
> I'd guess the tests failing might be causing it to exit before it gets there?

That's a thought. I can look into that next I'm able. 👍

PR #480, opened 2025-04-10
Enable erase-components flag for debug mode

benwis:
Enabling this flag by default will offer significant compile time decreases and resolve some issues with debug symbols in dev mode. Also includes the flags `disable_erase_components` to stop this behavior and `always_enable_erase_components` to do this for both debug and prod

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -533,7 +533,7 @@ dependencies = [
 
 [[package]]
 name = "cargo-leptos"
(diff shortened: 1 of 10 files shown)


gbj (approved):
Thanks, Ben! I misunderstood and broke it while trying to fix the tests, but un-broke it. Tested locally and it does correctly apply the setting. Just letting CI run.

PR #481, opened 2025-04-12
`cargo leptos new`: validate argument relationships

flo-l:
This PR closes #475.

It introduces two changes to the `cargo leptos new` subcommand:

- make either `--git` or `--path` required
- only allow `--branch` or `--tag` when `--git` is given

### How to test
`cargo run -- new --init` > should ask you to provide either `--git` or `--path`
`cargo run -- new --branch b` > should ask you to provide `--git`
`cargo run -- new --tag t` > should ask you to provide `--git`
`cargo run -- new --git g --path p` > should error with message explaining that these args are exclusive
`cargo run -- new --git g --tag t --branch b` > should error with message explaining that these args are exclusive

PR #482, opened 2025-04-12
remove unused import

flo-l:
I've seen this warning while working on #481 so I've decided to fix it while I have my editor open.


benwis:
I too have seen this warning, thanks for the fix

PR #489, opened 2025-04-25
Fix Cargo.lock being out of sync with Cargo.toml

sgued:
The latest release includes an out of date Cargo.lock, which prevents building with `--locked`

This PR adds a `cargo fetch --locked` step in the CI to prevent this from happening again in the future

Clos


benwis:
Closing as this is already done in the repo, but perhaps not in the released version

PR #499, opened 2025-05-02
fix: respect env rustflags, closes #496

### LINKED ISSUE #496: build fails due to `RUSTFLAGS` overwriting behavior
The current implementation overwrites user-defined `RUSTFLAGS` when enabling `erase_components`, which breaks builds that require specific RUSTFLAGS (such as `getrandom`'s `wasm_js` backend).

## Steps to Reproduce

Configure a project with `.cargo/config.toml` containing wasm-specific flags:

```toml
[target.wasm32-unknown-unknown]
rustflags = ["--cfg", "getrandom_backend=\"wasm_js\""]
```

## Current Behavior
wasm build fails with error, because `cargo-leptos` is overwriting `RUSTFLAGS` with just `--cfg erase_components`, discarding the user's configuration.

## Solution
Currently this could be bypassed by setting `disable-erase-components = true` and letting `cargo` apply the flags from `.cargo/config.toml`:

```toml
[target.wasm32-unknown-unknown]
rustflags = [
    "--cfg", "getrandom_backend=\"wasm_js\"",
    "--cfg", "erase_components"
]
```

What do you think is the best approach? Is there a way `cargo-leptos` could detect or preserve flags from .cargo/config.toml?

sabify:


benwis:
That seems like it'd work for env var RUSTFLAGs but probably not .cargo/config.toml RUSTFLAGS yes?

sabify:
Oops! AFAIK, there is no way you can set a rustflags without overriding it somewhere else.

@benwis I believe the changes should be reverted back; it interferes with the user's settings. Could you please do it and cut a release?

Note: In any case if cargo-leptos debug mode is used for examples in leptos, no matter the CI rule, erase_components will be used which is undesirable. I think the best way would be to add a section in README.md and educate people about enabling this flag.

PR #500, opened 2025-05-02
fix: respect env rustflags, closes #496

### LINKED ISSUE #496: build fails due to `RUSTFLAGS` overwriting behavior

sabify:
I found the trick. #499 follow-up (I closed it and couldn't reopen it).

@benwis This PR works as expected. Please check it out. The only concern about his feature is that we should make sure the leptos examples will not run on debug builds as it will  be always on `erase_components` (Not related to this PR).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -494,6 +494,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "cargo-config2"
+version = "0.1.32"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6dc3749a36e0423c991f1e7a3e4ab0c36a1f489658313db4b187d401d79cc461"
+dependencies = [
+ "serde",
+ "serde_derive",
+ "toml_edit",
+ "windows-sys 0.59.0",
+]
+
+[[package]]
 name = "cargo-generate"
 version = "0.23.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


benwis:
As far as I can tell my feat was unfounded. Thanks for patching this

luxalpa:
As a future side-note, this change also fixes cargo cranelift (`cargo-clif`) builds (which require rustflags to be set explicitly as it doesn't read them out by itself).

PR #504, opened 2025-05-05
Only return from build_proj after initially building all asset types. (Fixes #300)

### LINKED ISSUE #300: `watch` does not load style assets if the initial build fails
1. Run cargo leptos watch in your project while there's some code error
2. Fix that error
3. The app that now loads doesn't have any CSS until you restart the watch.

This has been a surprisingly annoying issue for me. I think it's caused by https://github.com/leptos-rs/cargo-leptos/blob/d36a1870452fceab2365df6ae16677b6573b69a3/src/command/build.rs#L38-L50 where it just returns and doesn't load any assets if the build fails. Further style changes are only added iteratively.

I assume we can just run the `compile::assets` and `compile::styles` functions (and possibly others) regardless of success or failure. Maybe I'll drop a PR for this later.

luxalpa:
In the muture we might also want to tackle the `add_hashes_to_site` case and the case of `build` vs `watch`, but for now I think this is sufficient, and having used this for a couple of months I haven't seen any problems.

Fixes #300

### DIFF
--- a/src/command/build.rs
+++ b/src/command/build.rs
@@ -33,13 +33,19 @@ pub async fn build_proj(proj: &Arc<Project>) -> Result<bool> {
     }
     let changes = ChangeSet::all_changes();
 
+    let mut success = true;
+
     if !compile::front(proj, &changes).await.await??.is_success() {
-        return Ok(false);
+        success = false;
     }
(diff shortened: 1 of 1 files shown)


benwis:
Seems reasonable, sorry that's been bugging you. Thanks for the PR

PR #524, opened 2025-05-21
Fix name for brackets in README.md

dbdr:
I believe "braces" always refer to { } characters. The name for [ ] is brackets (american english) or square brackets (british english). Using the square form here makes it unambiguous.

Source: https://en.wikipedia.org/wiki/Bracket


benwis:
Thanks for the PR

PR #525, opened 2025-06-05
chore: update tailwind version

g1ibby:


src/config/version.rs
```diff
@@ -17,7 +17,7 @@ impl VersionConfig {
 
     pub fn default_version(&self) -> &'static str {
         match self {
-            Self::Tailwind => "v4.1.4",
+            Self::Tailwind => "v4.1.8",
```
cramhead, line 20: I've been using the v4.1.10 version. Maybe that would be avoid a change after this one goes through.
```suggestion
            Self::Tailwind => "v4.1.10",
```

gbj (approved):

Issue #526, opened 2025-06-10
Release builds broken on Rust 1.87


paul-hansen:
After updating to Rust 1.87, release builds are failing with cargo-leptos at the wasm-opt step.
```
Error: 
   0: at `/home/runner/work/cargo-leptos/cargo-leptos/src/compile/front.rs:53:30`
   1: at `src/compile/front.rs:203:10`
   2: Failed to read module
   3: [parse exception: invalid code after misc prefix: 17 (at 0:4096996)]

Location:
   src/compile/front.rs:203

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
Error: Process completed with exit code 1.
```
leptos 0.6
cargo-leptos 0.2.35

Appears to be a relevant comment here: https://github.com/rustwasm/wasm-pack/issues/1441
> Rust 1.87.0 moved to LLVM 20, which enabled more webassembly features by default.
>
> wasm-opt needs additional commandline flags to enable the same webassembly features, or it will fail with this error. See earlier comments for some flags to try.

The update to llvm 20 is mentioned at the bottom of the release notes here: https://github.com/rust-lang/rust/releases/tag/1.87.0
The LLVM 20 release notes related to webassembley are here: https://releases.llvm.org/20.1.0/docs/ReleaseNotes.html#changes-to-the-webassembly-backend

jarverr:
```
Front generating JS/WASM with wasm-bindgen
Finished generating wasm-bindgen output in 109.707106ms
Finished emitting wasm-bindgen in 135.897847ms

Error:
   0: at `/home/runner/work/cargo-leptos/cargo-leptos/src/compile/front.rs:53:30`
   1: at `src/compile/front.rs:203:10`
   2: Failed to read module
   3: [parse exception: invalid code after misc prefix: 17 (at 0:519416)]

Location:
   src/compile/front.rs:203
```

Got the same issue when trying to build for release on:
leptos 0.8.2
cargo-leptos 0.2.35
rust: 1.89.0-nightly

paul-hansen:
Looks like Dioxus got around this by switching back to only using the wasm-opt binary: https://github.com/DioxusLabs/dioxus/pull/4247

[...]

PR #527, opened 2025-06-10
Attempt to fix release builds failing with wasm-opt

paul-hansen:
See #526

Enables the two features that LLVM 20 enabled to try to fix the release builds failing at wasm-opt with Rust 1.87.

Seems like this should have fixed it but I tried it and I get the same error (just with the line number pushed down two lines.


paul-hansen:
Closing in favor of #534

Issue #532, opened 2025-06-16
`.gitignore` takes priority over `watch-additional-files` option in `watch` command


AkinAguda:
I want to watch a file is in my .gitignore and adding it to the `watch-additional-files` array in my `Cargo.toml` does not seem to work

My use case:
I am combining [stylance](https://github.com/basro/stylance-rs) with tailwind. And my style folder looks like this:
<img width="279" alt="Image" src="https://github.com/user-attachments/assets/3efafacc-451a-4401-b5b6-202f0c963941" />

When stylance compiles all my modular css and puts it in the `compiled.scss` file, I want `cargo-leptos` to watch that file for changes. I also need it to be in my gitignore since its compiled CSS and what really should be pushed are the source modular CSS.

These are my configurations:

<img width="416" alt="Image" src="https://github.com/user-attachments/assets/bc6c15ac-bccf-4c9c-82ef-ce66e7913cd0" />
<img width="310" alt="Image" src="https://github.com/user-attachments/assets/d800a3b0-3b6e-4860-8f28-8770e1e912e8" />

`cargo-leptos` only watches then I remove this line from my gitignore

<img width="181" alt="Image" src="https://github.com/user-attachments/assets/91dcdfe8-7fe8-4975-9b84-28e4ca37a99b" />

PR #533, opened 2025-06-16
fix: `watch-additional-files` should now override `.gitignore`

AkinAguda:
This PR addresses the issue I opened [here](https://github.com/leptos-rs/cargo-leptos/issues/532)


benwis:
Thanks for the PR, I'm collecting fixes for a new release so it should be available soon

PR #537, opened 2025-06-29
Add environment variable logging before executed commands

mahdi739:


mahdi739:
I'm closing this PR because the environment variables are already being logged at the debug level. (with `-v` flag)

PR #539, opened 2025-07-02
fix: mismatch between aarch64 and arm64 naming on MacOS

saikatdas0790:


src/ext/exe.rs
```diff
@@ -508,42 +508,27 @@ impl Command for CommandWasmOpt {
...
-            ("linux", "aarch64") => Ok(format!(
-                "https://github.com/{}/{}/releases/download/{}/binaryen-{}-arm64-linux.tar.gz",
-                self.github_owner(),
-                self.github_repo(),
-                version,
-                version
-            )),
+            ("windows", "x86_64") => Ok(format!("{base_url}-x86_64-windows.tar.gz")),
+            ("macos", "x86_64") => Ok(format!("{base_url}-x86_64-macos.tar.gz")),
+            ("macos", "aarch64") => Ok(format!("{base_url}-arm64-macos.tar.gz")),
```
sabify, line 528: @saikatdas0790 Can you please also add the following line as a future proof correct targets?

```rust
("macos", "arm64") => Ok(format!("{base_url}-arm64-macos.tar.gz")),
```

saikatdas0790:
@sabify Added the "macos" <-> "arm64" combo to the list as well

gbj (approved):
Thank you!

PR #543, opened 2025-07-10
Enable nontrapping-fptoint feature while invoking wasm-opt

godvino:
Without enabling this feature, trying to build a  Leptos project in release profile with strip enabled results in errors such as:
```    wasm-opt failed with:
[wasm-validator error in function 27] unexpected false: all used features should be allowed, on 
(i32.trunc_sat_f64_u
 (local.get $8)
)
[wasm-validator error in function 27] unexpected false: all used features should be allowed, on 
(i32.trunc_sat_f64_u
 (local.get $9)
)
[wasm-validator error in function 207] unexpected false: all used features should be allowed, on 
(i32.trunc_sat_f64_u
 (local.get $4)
)
Fatal: error validating input
```
See https://github.com/rust-lang/rust/pull/137322

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,9 +230,9 @@ disable-erase-components = false
 
 # Enabling wasm-opt features
 #
-# Optional. By default, "-Oz" and "--enable-bulk-memory" are used. For all features, consult `wasm-opt --help`.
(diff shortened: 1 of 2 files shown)


gbj (approved):
Looks good to me, thank you!

PR #550, opened 2025-07-17
Implement --frontend-only and --backend-only options for build commands

wrl:
My PaaS prefers to have backends built in their own CI runners, but they're slow, so this lets me build the frontend in my own (faster) CI and then push the generated `target/site` along with the rest of the codebase into their CI.


gbj (approved):
Fine by me!

wrl:
> Fine by me!

@gbj, thanks! it seems like i need explicit approval to re-run CI for the test fixes.

PR #554, opened 2025-07-29
Add shell completions to `cargo-leptos`

titaniumtraveler:
Since `cargo` now supports forwarding of shell completions of `cargo
<custom-cmd>` to `cargo-<custom-cmd>` for at least `bash` and `zsh`,
providing the backing completions seems useful.

They still have to be manually installed by the user, for `bash` that can be
done by either running `eval "$(cargo leptos completions bash)"` to enable the
completions for the *current* session and

```bash
cargo leptos completions bash > "${XDG_DATA_HOME:-"$HOME/.local/share}/bash-completion/completions/cargo-leptos"
```

to let `bash` lazily load the completions as soon as they are needed.

For `zsh` I'm not quite sure how they are working, but considering how other
projects manage autocompletions, describing the details of how the shells
completions work, might be not really necessary *anyway*.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -596,6 +596,7 @@ dependencies = [
  "cargo-generate",
  "cargo_metadata",
  "clap",
+ "clap_complete",
  "color-eyre",
  "derive_more",
  "dirs",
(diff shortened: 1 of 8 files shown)


benwis:
Thanks for the PR!

PR #559, opened 2025-08-13
fix: truncate long filenames (closes #556)

### LINKED ISSUE #556: File name too long for wasm chunk
This produces names that might be way above the 255 bytes of filesystem limits with 10-15 lazy routes:

https://github.com/leptos-rs/cargo-leptos/blob/b8883fbe007c4b180b3d7054284f111f908b55ac/src/wasm_split_tools/split_point.rs#L253-L267

Changing it to something like `if splits.join("_").len() >= 200 { hash(splits.join("_")) } else { splits.join("_") }` seems to have fixed it for me, but not entirely sure, still figuring out a couple other issues with wasm splitting that may or may not be caused by this change

gbj:
@Sliman4 I don't have a way to reproduce issue #556 but this seems to me to be a reasonable way to truncate long filenames. Does it fix your issue?

### DIFF
--- a/src/wasm_split_tools/split_point.rs
+++ b/src/wasm_split_tools/split_point.rs
@@ -258,8 +258,8 @@ impl SplitModuleIdentifier {
             Self::Split { name, .. } => name.clone(),
             Self::Chunk { splits, hash } => {
                 let mut full_name = splits.join("_");
-                if full_name.len() > 200 {
-                    full_name.truncate(200);
+                if full_name.len() > 128 {
+                    full_name.truncate(128);
                     full_name.push_str(hash);
 
                     let mut hasher = DefaultHasher::new();
@@ -279,8 +279,8 @@ impl SplitModuleIdentifier {
             Self::Split { name, hash } => format!("{name}.{hash}"),
             Self::Chunk { splits, hash } => {
                 let mut splits = splits.join("_");
-                if splits.len() > 200 {
-                    splits.truncate(200);
+                if splits.len() > 128 {
+                    splits.truncate(128);
                 }
                 format!("{splits}.{hash}")
             }


gbj:
See also https://github.com/leptos-rs/leptos/issues/4223

Sliman4:
It does seem to fix the issue in debug mode, but in release now I'm getting this:

Uncaught (in promise) SyntaxError: redeclaration of const __wasm_split_load_a_view_11497074213207055331_b_view_11497074213207055331_c_view_11497074213207055331_d_view_JroW7bMxjD__lKieADD0CA __wasm_split.______________________.js:149:7
note: Previously declared at line 148, column 7

I guess hash doesn't necessarily make the file name unique or something, didn't read the code where it takes the hash from. Though not sure if it's a different issue or related to this. But this commit fixed it for me, so maybe hashing the splits provides more uniqueness https://github.com/Sliman4/cargo-leptos-temp-fix/commit/fa9da34380de59d87124924c2fc762693e120d88

benwis:
Perhaps it would be better to generate some uuids here to prevent conflics?

gbj:
@Sliman4 I don't think you've provided a way to reproduce the issue, so I can't test directly, but could you try testing your example in release mode against the latest version of this PR?

Sliman4:
Now getting this error

```
   0: at `/home/slime/.cargo/git/checkouts/cargo-leptos-929e2b17adbc75b2/2d1cd72/src/command/serve.rs:8:46`
   1: at `src/compile/hash.rs:18:56`
   2: Failed to rename dist/pkg/<200 chars>_11176612011265763653.F714K2b9n4VLlJAM0gUWkQ.wasm to dist/pkg/<200 chars>_11176612011265763653.F714K2b9n4VLlJAM0gUWkQ.b4kR1cTL5olEeYD2fe-Y7A.wasm at `src/compile/hash.rs:187:14`
```

The first filename is 249 characters, the other is 272 which is too long. I think there's no need to increase the substring length to 200 if there's a hash of the entire string that already includes the uniqueness of the entire string

I currently don't have time to make a small reproducible example, but I guess creating ~20 lazy routes should be able to reproduce this

gbj:
The other test case I have is an example with 17 lazy routes and fairly long names, and it's fine on all these commits. I've reduced it back down to 128. If you could test again I'd appreciate it.

(Apologies for the ping-pong, I just don't want to merge something that doesn't fix the problem for you, but I don't have a way to reproduce the particulars so it's a bit of guesswork...)

Sliman4:
When I load a lazy route, some of the requests succeed, but some fail because the browser makes a request to `/pkg/account_settings_view_18364835182258006813_connected_apps_settings_view_18364835182258006813_preferences_settings_view_183648351.WMMAM5A4qAjuwe-AZ0isWQ.wasm`, but in my pkg folder it's actually named `dist/pkg/account_settings_view_18364835182258006813_connected_apps_settings_view_18364835182258006813_preferences_settings_view_183648351_8831730838953392598.WMMAM5A4qAjuwe-AZ0isWQ.WMMAM5A4qAjuwe-AZ0isWQ.wasm` (has many more different hash suffixes). Probably caused by differences in `name` and `name_hashed` implementations.

I actually have a working fix on https://github.com/Sliman4/cargo-leptos-temp-fix/commit/fa9da34380de59d87124924c2fc762693e120d88 (same implementation as on suggested fix in the original issue), was a bit hesitant to make a PR because I didn't know if this needs to be changed in any other places or if it could break anything because I'm not particularly fluent in this codebase, but since your fix affected only these 2 methods as well, maybe it would be easier for me to make a PR with the changes? (and replace "too_long_" with maybe a trimmed original name)

gbj:
If you have a fix that works consistently for your case, please make a PR and I will test it against the cases for which I do have access to a reproduction!

Issue #557, opened 2025-08-12
Panic when using --split on second invocation with no changes


lukexor:
I'm having this issue in my main project, and get one error, but was able to get a different error using the lazy_routes example so I'm not sure if they're related or not.

I'm running `cargo leptos watch -p <package> --split --wasm-debug`. It runs fine on a clean build. Once it's started, if you Ctrl+C and run it again, you get the following error:

```
❯ cargo leptos watch -p <package> --split --wasm-debug -- start
    Finished `dev` profile [optimized] target(s) in 0.11s
       Cargo finished cargo build --package=<package> --lib --target-dir=/home/luke/.cargo-target/front --target=wasm32-unknown-unknown --no-default-features --features=hydrate
       Front splitting out lazy-loaded WASM files
    Finished WASM splitting in 12.033833ms
       Front generating JS/WASM with wasm-bindgen
Error:
   0: at `/home/runner/work/cargo-leptos/cargo-leptos/src/compile/front.rs:71:38`
   1: at `/home/runner/work/cargo-leptos/cargo-leptos/src/compile/front.rs:151:10`
   2: failed getting Wasm module for '../../.cargo-target/front/wasm32-unknown-unknown/debug/<package>.wasm'

Location:
   /home/runner/work/cargo-leptos/cargo-leptos/src/compile/front.rs:151

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
                                 ⋮ 1 frame hidden ⋮
   2: <E as eyre::context::ext::StdError>::ext_report::hb0a0e09da9b4dfc4
      at <unknown source file>:<unknown line>
   3: cargo_leptos::compile::front::bindgen::{{closure}}::hc6d5df8c66f3b36a
      at <unknown source file>:<unknown line>
   4: cargo_leptos::compile::front::front::{{closure}}::{{closure}}::had6a99af2640228d
      at <unknown source file>:<unknown line>
   5: tokio::runtime::task::raw::poll::h6426209ce24a1e68
      at <unknown source file>:<unknown line>
   6: tokio::runtime::scheduler::multi_thread::worker::Context::run_task::h429df9855732d72a
      at <unknown source file>:<unknown line>
   7: tokio::runtime::scheduler::multi_thread::worker::run::he31d8fe9a42607d0
      at <unknown source file>:<unknown line>
[...]

PR #563, opened 2025-09-04
fix: properly handle split chunk names

sabify:
Supersedes:  #559
Closes: #556
Closes: #557

This PR properly handles the split chunk names.

JS variable names will be human-readable and debug-friendly and the module names will be hashed and loaded with their hashes.

### DIFF
--- a/src/wasm_split_tools/mod.rs
+++ b/src/wasm_split_tools/mod.rs
@@ -118,40 +118,35 @@ function makeLoad(url, deps) {
     );
     let mut split_deps = HashMap::<String, Vec<String>>::new();
     let mut manifest = HashMap::<String, Vec<String>>::new();
-    for (name, _) in split_program_info.output_modules.iter() {
-        let SplitModuleIdentifier::Chunk { splits, .. } = name else {
+    for (identifier, _) in split_program_info.output_modules.iter() {
+        let SplitModuleIdentifier::Chunk { splits, hash } = identifier else {
             continue;
         };
-        let chunk_name = name.name(proj);
-        let chunk_hash = name.hash(proj);
+        let name = identifier.name(proj);
 
-        manifest
-            .entry(chunk_name.clone())
-            .or_default()
-            .push(chunk_hash.clone());
+        manifest.entry(name.clone()).or_default().push(hash.clone());
 
         for split in splits {
             split_deps
                 .entry(split.clone())
                 .or_default()
-                .push(chunk_name.clone());
+                .push(name.clone());
             manifest
                 .entry(split.clone())
                 .or_default()
-                .push(chunk_hash.clone());
+                .push(hash.clone());
         }
 
         javascript.push_str(
             format!(
-                "const __wasm_split_load_{chunk_name} = makeLoad(new URL(\"./{chunk_hash}.wasm\", import.meta.url), []);\n",
+                "const __wasm_split_load_{name} = makeLoad(new URL(\"./{hash}.wasm\", import.meta.url), []);\n",
             ).as_str()
         );
     }
     for (identifier, _) in split_program_info.output_modules.iter().rev() {
-        let SplitModuleIdentifier::Split { name, .. } = identifier else {
+        let SplitModuleIdentifier::Split { name, hash } = identifier else {
             continue;
         };
-        let hash = identifier.hash(proj);
 
         manifest
             .entry(identifier.name(proj))


sabify:
@Sliman4 Could you please test this PR against your use case?

sabify:
This also fixes the issue #557. Please see the comment https://github.com/leptos-rs/cargo-leptos/issues/557#issuecomment-3262881283.

gbj:
This seems okay to me. It certainly fixes any issues with filename length. And using the `_split` name is a good idea due to #557. 

I do think that having *some* human-readable indicator in the filename is useful, even just for helping people develop intuitions about what is loading when while they're learning about lazy-loading in general.

Here's the before/after in devtools

https://github.com/user-attachments/assets/3ecf7b72-1f81-4299-9c61-4d402af14bd6

https://github.com/user-attachments/assets/1ac012bc-2eea-41c8-8f20-0a70fec5533b

I wonder if something as simple as `name_hashed` using a `truncated_name.{hash}` a `Split` and `shared.{hash}` for a chunk would help.

I understand that they can be looked up in the table in `__wasm_split.js`, even if you only have the hashes... In this comment I am *only* thinking about the developer-experience-and-education benefit of having some developer-readable filenames visible in the network tab.

Sliman4:
> @Sliman4 Could you please test this PR against your use case?

My browser makes a request to `/gICpufDiiUwMp-JmEl6mSA.gICpufDiiUwMp-JmEl6mSA.gICpufDiiUwMp-JmEl6mSA.wasm` but I only have `gICpufDiiUwMp-JmEl6mSA.gICpufDiiUwMp-JmEl6mSA.wasm`, so it results in a 404. It seems the hash is repeated either 2 or 3 times for some reason. Same is happening for all lazily loaded .wasm files in my project. When I try running the lazy_routes example, the hash is not duplicated, it's just `http://127.0.0.1:3000/pkg/j7z39lLeUs7H6np9sdrA4g.wasm`

Update: Found that it's caused by `hash-files = true` in `[[workspace.metadata.leptos]]` in Cargo.toml. Without hash-files, the hash is not duplicated and is loaded correctly.

sabify:
> This seems okay to me. It certainly fixes any issues with filename length. And using the `_split` name is a good idea due to #557.
> 
> I do think that having _some_ human-readable indicator in the filename is useful, even just for helping people develop intuitions about what is loading when while they're learning about lazy-loading in general.
> 
> Here's the before/after in devtools
> Screen.Recording.2025-09-07.at.8.33.43.PM.mov
> Screen.Recording.2025-09-07.at.8.34.33.PM.mov
> 
> I wonder if something as simple as `name_hashed` using a `truncated_name.{hash}` a `Split` and `shared.{hash}` for a chunk would help.
> 
> I understand that they can be looked up in the table in `__wasm_split.js`, even if you only have the hashes... In this comment I am _only_ thinking about the developer-experience-and-education benefit of having some developer-readable filenames visible in the network tab.

By saying it is human-readable and debug-friendly, I mean it is useful for maintainers of this tool and leptos itself to debug the api.

Users of this tool may not benefit from the naming and only seeing the applied logic in action would be sufficient for them. (Like seeing the modules are loading on request... and I think more than that wouldn't be informational)

sabify:
> > @Sliman4 Could you please test this PR against your use case?
> 
> My browser makes a request to `/gICpufDiiUwMp-JmEl6mSA.gICpufDiiUwMp-JmEl6mSA.gICpufDiiUwMp-JmEl6mSA.wasm` but I only have `gICpufDiiUwMp-JmEl6mSA.gICpufDiiUwMp-JmEl6mSA.wasm`, so it results in a 404. It seems the hash is repeated either 2 or 3 times for some reason. Same is happening for all lazily loaded .wasm files in my project. When I try running the lazy_routes example, the hash is not duplicated, it's just `http://127.0.0.1:3000/pkg/j7z39lLeUs7H6np9sdrA4g.wasm`
> 
> Update: Found that it's caused by `hash-files = true` in `[[workspace.metadata.leptos]]` in Cargo.toml. Without hash-files, the hash is not duplicated and is loaded correctly.

Please checkout the last commit, it solves the issue with double hashing of split file names with `hash_files = true`.

benwis:
Love to see it, thanks for the PR @sabify

Sliman4:
> Please checkout the last commit, it solves the issue with double hashing of split file names with `hash_files = true`.

Yes, it solved the issue, now everything is loading correctly

Issue #574, opened 2025-10-01
split generates wrong wasm files


johansmitsnl:
When you run `cargo leptos build --split` is generates html where it requests `pkg/web-app_bg.wasm`. But if you look in the `pkg` dir there is no such file with the extention `_bg.wasm`. Only `pkg/web-app.wasm`.

Is this just a naming error of the file of is the `_bg` a  different file that was not generated?

Now the UI is not functional due to the load failure.

johansmitsnl:
It looks like the [rename](https://github.com/leptos-rs/cargo-leptos/blob/e7c09a9a456060d3f3542590726832f6190d7f84/src/compile/front.rs#L178-L189) is not working?

Edit, I also found this in leptos [hydration](https://github.com/leptos-rs/leptos/blob/bbf5bf9170aa862053619f5c939fc885beaa7e99/leptos/src/hydration/mod.rs#L155-L156) pushing the name with the `_bg` in it if the env `LEPTOS_OUTPUT_NAME` was not set.

gbj:
These issues (this and https://github.com/leptos-rs/leptos/issues/4347) give me the impression that there is something happening in your setup that I cannot reproduce and that other users have not encountered. 

I'm also a little unclear on what you mean: `cargo leptos build --split` doesn't generate HTML at all, so I'm not sure how it would generate HTML that requests the wrong WASM file.

Could you provide some steps to reproduce the problem? Because it does not occur in the general case.

johansmitsnl:
I'm trying to debug what was happening (still in progress but a state update), this is what I found until now:

1. `cargo leptos build --split` builds the wasm file `pkg/web-app_bg.wasm` but does a [rename](https://github.com/leptos-rs/cargo-leptos/blob/e7c09a9a456060d3f3542590726832f6190d7f84/src/compile/front.rs#L178-L189) to  `pkg/web-app.wasm` for backwards compatibility. I don't know (lack of knowledge) what this backwards compatibility is. To me it should be if the `LEPTOS_OUTPUT_NAME` env variable was not set. See point 5.
2. The binary is shipped to another location (away from finding the `Cargo.toml`)
3. The leptos_config is initialized from `leptos::config::get_config_from_env()`
4. Running the app in a different location requires a `LEPTOS_OUTPUT_NAME=web-app` env variable to be set.
[...]

PR #576, opened 2025-10-02
Add documentation for cross (Closes: #574)

johansmitsnl:
When using cross not all variables are passed into the build environment. This documents how to do this with cross but it might also be relevant to other build systems.


gbj:
Thanks!

PR #578, opened 2025-10-07
Replace vendored wasm-split with out-of-repository version

WorldSEnder:
PR is for testing for now. The git dependency must be replaced by a proper release of the `wasm-split` crate before merging.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -630,6 +630,7 @@ dependencies = [
  "walkdir",
  "wasm-bindgen-cli-support",
  "wasm-encoder 0.235.0",
+ "wasm_split_cli_support",
  "wasmparser 0.235.0",
  "which",
  "zip",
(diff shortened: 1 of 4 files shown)


WorldSEnder:
Sounds good, name squat is unfortunate.

I wanted to proceed with giving it a few more tests before releasing and am currently battling wasm-bindgen-test-runner a bit for that. There are some corners where the tool makes assumptions about llvm output that are not super explicit in the code and I didn't want to reintroduce some regressions by rewriting. With your blessing though, I could imagine releasing a "0.2.0" with data splitting, then work on tests and then build up a 0.2.x with the smaller stuff ironed out.

There is also an idea of a better graph walking algorithm in my mind (by first building connected components and dominators, then assigning those to modules/chunks/main instead of individual functions with a full depth-first traversal for each split which should be faster. Dioxus does similar things. Also an improved chunking, current one has the downside of exponential explosion in the number of splits when they pairwise share common code, I think?). When I have sufficient time for that anyway, hence testing first :)

Finally, the increase in total size is something that deserves a second look. Might be connected to overlapping data symbols, llvm can merge immutable data like strings to overlap in memory and the current implementation copies them into nonoverlapping chunks - which is okay when they end up in different modules, but should probably get merged again when they end up in the same.

johansmitsnl:
One question I have, by having the files named `chunk_xx` would this not be a issue with caching with cdn's? Currently the splitted files have some hash method. When releasing a new version and not have to deal with cache headers/cdn settings would a hash based on the content be better (even without the leptos hash function)?

WorldSEnder:
I'm happy to put the hashing back in if it serves a use. I mainly interpreted it as a nice way to not have to deal with a potentially very long file name compared with the original prototype. Shouldn't a correctly working CDN (with default settings) already serve an etag that let's the client figure out that a new version must be fetched? Might be me being overly confident about the "Real World" and not having experienced badly behaving proxies and caches in a while.

gbj:
I agree that `chunk_X` is fine -- correct that the hashed names were a way to deal with file name length. `cargo-leptos` does add the hashes to these files correctly if you have the file-hashing setting enabled, so the behavior is consistent with all other files it produces.

WorldSEnder:
Quick update: I'm confident with it integrating into yew just the same now. Did some work on making the crate easier to semver bump upwards. I think the only remaining open item is worrying about reference types which are hackily removed before `wasm-bindgen` gets the chance to bungle things up. Doesn't need to happen in this release though, so next up is finishing some readme and documentation, then release the current `rc` as `0.2.0`. I'd let this pend for a few days if you want to do some final testing with the current `rc`.

gbj:
Oh wow this has gotten even better than the last time I tested. I do not seen any issues... It also appears that it's WASM-binary compatible with the current published Leptos version, which is awesome. Of course I will merge the `leptos` crate PR as well to migrate to the share helpers/macro crates, but it looks like the updated `cargo-leptos` won't require the updated `leptos` to use the new WASM-splitting, which is great. 

I'll let this sit while you finish up the polish, and will merge and publish a new version once you've release 0.2.0.

Issue #479, opened 2025-04-09
`cargo install --locked cargo-leptos` requires `perl`


splurf:
Attempting to install `cargo-leptos` with the provided command. The compilation process stops around halfway if `perl` isn't installed.

--- stderr
    Error configuring OpenSSL build:
        Command 'perl' not found. Is perl installed?
        ...

Steps to reproduce:
- Be on a Windows 10 machine.
- Have Perl not installed.
- `cargo install --locked cargo-leptos`

benwis:
Unfortunately this isn't a bug but a side-effect of the build process required to generate binaries for different platforms. Fixing that or porting us to rusttls would be a welcome PR, but I'm closing this issue for now

splurf:
> Unfortunately this isn't a bug but a side-effect of the build process required to generate binaries for different platforms. Fixing that or porting us to rusttls would be a welcome PR, but I'm closing this issue for now

Figured `perl` could be listed in the README.md as a dependency.

MaddyGuthridge:
Notably, the specific dependencies you need vary depending on your OS. For me, I needed `pkgconf` `perl-FindBin` `perl-IPC-Cmd` and `openssl-devel`

I suggest we link to the documentation on this, so people can determine which tools they need: https://docs.rs/openssl/latest/openssl/#automatic

Although porting to something that is a less of a nightmare to compile for sounds like a good idea.

gbj:
Cross-posting from my comment on #580:

> Just for anyone reading who's curious: it looks like OpenSSL is currently used only as a dependency of cargo-generate (which we use for cargo leptos new). https://github.com/cargo-generate/cargo-generate/issues/1008 discusses the possibility of using rustls in cargo-generate, which then links to the relevant issue in git2 (https://github.com/rust-lang/git2-rs/issues/623) — once cargo-generate can migrate to gitoxide instead, this will apparently also be possible, but that is not yet fully supported.

PR #580, opened 2025-10-09
Document openssl-sys dependency requirements

MaddyGuthridge:
I, like many other users, ran into issues installing `cargo-leptos` due to struggles getting its `openssl-sys` dependency to build.

As per @benwis 's comment on #479, it'd be nice to switch to rusttls, which would probably be easier to compile, but in the mean-time, updating the readme to warn future users of this potential pain-point will help future users resolve it themselves. As such, I have added a link to the `openssl` crate's documentation, which specifies the dependencies required for compilation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,11 @@ Install:
 
 > `cargo install --locked cargo-leptos`
 
+Note that this may require additional dependencies for SSL on some systems.
+If the installation fails while building `openssl-sys`, you will need to
+[install additional tools](https://docs.rs/openssl/latest/openssl/#automatic),
+before retrying the build.
+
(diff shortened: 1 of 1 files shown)


gbj:
Thanks for adding the docs link.

Just for anyone reading who's curious: it looks like OpenSSL is currently used only as a dependency of `cargo-generate` (which we use for `cargo leptos new`). https://github.com/cargo-generate/cargo-generate/issues/1008 discusses the possibility of using `rustls` in `cargo-generate`, which then links to the relevant issue in `git2` (https://github.com/rust-lang/git2-rs/issues/623) — once `cargo-generate` can migrate to `gitoxide` instead, this will apparently also be possible, but that is not yet fully supported.

PR #582, opened 2025-10-17
change: use external `wasm-bindgen` CLI rather than compiling it into the binary

gbj:
Traditionally, `cargo-leptos` has compiled `wasm-bindgen-cli-support` directly into the binary. This is nice because it allows us to avoid downloading and installing an extra binary, and because it allows us to directly use the results of `wasm-bindgen` within the `cargo-leptos` process and emit them however we want, including minifying JS.

On the other hand it is a source of perpetual confusion. `wasm-bindgen-cli-support` is intentionally not intended to be used by anything other than `wasm-bindgen-cli`, and it 1) breaks semver by requiring an updated `wasm-bindgen-cli-support` patch version to match every updated `wasm-bindgen` patch version and 2) emits baked-in error messages to stderr if those don't match that tell you to update `wasm-bindgen`, which does not work (because it's `cargo-leptos` that needs to be updated instead).

I'm not interested in continuing to patch `cargo-leptos` to the latest version at random times when `wasm-bindgen` is updated, or in trying to coach users through the confusing error messages.

This PR changes the behavior to download and use `wasm-bindgen` just like we do other tools (SASS, wasm-opt, and Tailwind).

This is a big change internally, but should not affect users. However, I'd really appreciate some help testing it on different platforms and with various build configurations.

Edit to add: You can test this by installing with `cargo install --git https://github.com/leptos-rs/cargo-leptos --branch separate-wasm-bindgen-cli cargo-leptos` and then using `cargo-leptos` and seeing if I've broken anything.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,4 +1,3 @@
-
 name: Test
 
 on: [push, pull_request]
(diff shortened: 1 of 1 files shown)


fgardt:
Tested this change on my current leptos project and it seems to work fine. Tho its a rather simple application.

Issue #588, opened 2025-11-25
[0.3.0] Determine `wasm-bindgen` version by inspecting project lock file?


LeoniePhiline:
We are seeing this failure in our CI since #582 

<details><summary>Details</summary>

```
      Front generating JS/WASM with wasm-bindgen
     Command [wasm-bindgen] requested version 0.2.104, but a newer version 0.2.105 is available, you can try it out by setting the LEPTOS_WASM_BINDGEN_VERSION=0.2.105 env var and re-running the command
     Command installing wasm-bindgen-0.2.104 ...
     Command wasm-bindgen-0.2.104 installed.
wasm-bindgen failed with:
[2025-11-24T01:29:53Z DEBUG walrus::module::types] parsing type section
[2025-11-24T01:29:53Z DEBUG walrus::module::imports] parse import section
[2025-11-24T01:29:53Z DEBUG walrus::module::functions] parse function section
[2025-11-24T01:29:53Z DEBUG walrus::module::tables] parse table section
[2025-11-24T01:29:53Z DEBUG walrus::module::memories] parse memory section
[2025-11-24T01:29:53Z DEBUG walrus::module::globals] parse global section
[2025-11-24T01:29:53Z DEBUG walrus::module::exports] parse export section
[2025-11-24T01:29:53Z DEBUG walrus::module::elements] parse element section
[2025-11-24T01:29:53Z DEBUG walrus::module::data] parse data section
[2025-11-24T01:29:53Z DEBUG walrus::module] parsing custom section `.debug_abbrev`
[2025-11-24T01:29:53Z DEBUG walrus::module] parsing custom section `.debug_info`
[2025-11-24T01:29:53Z DEBUG walrus::module] parsing custom section `.debug_ranges`
[2025-11-24T01:29:53Z DEBUG walrus::module] parsing custom section `.debug_str`
[2025-11-24T01:29:53Z DEBUG walrus::module] parsing custom section `.debug_line`
[2025-11-24T01:29:53Z DEBUG walrus::module] parsing custom section `__wasm_bindgen_unstable`
[2025-11-24T01:29:53Z DEBUG walrus::module] parsing custom section `__wasm_split_unstable`
[2025-11-24T01:29:53Z DEBUG walrus::module] parse name section
[2025-11-24T01:29:53Z DEBUG walrus::module::producers] parse producers section
[2025-11-24T01:29:53Z DEBUG walrus::module] parsing custom section `target_features`
[...]

PR #589, opened 2025-11-25
feat: detect wasm-bindgen version from project Cargo.lock (#588)

LeoniePhiline:
This MR implements #588:

Automatically detects and uses the `wasm-bindgen` version from the project's `Cargo.lock` file instead of immediately falling back to a hard-coded default version. This ensures version compatibility between the `wasm-bindgen` dependency and the `wasm-bindgen-cli` tool.

**Edit: I provided an alternate approach (with I prefer) as https://github.com/leptos-rs/cargo-leptos/pull/590.** Details of the alternate approach are discussed further down below.

The `LEPTOS_WASM_BINDGEN_VERSION` variable still has the highest priority, to not break the public API.

**Version resolution priority:**

1. `LEPTOS_WASM_BINDGEN_VERSION` environment variable (highest)
2. Version from project's `Cargo.lock`
3. Hard-coded default version (`0.2.104`)

**Implementation details:**

- Adds [`cargo-lock`](https://crates.io/crates/cargo-lock) dependency for parsing `Cargo.lock` files. (Solid dependency from [`rustsec`](https://github.com/rustsec/rustsec/tree/main/cargo-lock).)
- Creates `detect_wasm_bindgen_from_lockfile()` helper function.
- Updates `Exe::get()` to accept optional project root path parameter.
- Threads `project_root` through `Command` trait methods.

**Design decision:**

The project root path is passed as an optional parameter to `Exe::get()` rather than making `WasmBindgen` a tuple struct variant with an embedded path. This avoids adding lifetime parameters to the `Exe` enum and keeps the API simpler, at the cost of requiring `None` for tools that don't need project context.

This approach is less involved, but not entirely aesthetic. **I am hoping for your input here:**

If you prefer, I can instead provide a solution with `Exe<'a>::WasmBindgen(&'a Utf8Path)` and `struct CommandWasmBindgen<'a>(&'a Utf8Path)` (or alternatively a named struct field `project_root`, rather than tuple structs &mdash; let me know what you prefer).

This way, the root path would not need to be passed through `Exe::get` and `Exe::get_meta`, as well `Command::exe_meta` (which, as `trait` `fn`, requires the other commands to also accept the project root parameter). 

Instead, `Exe::meta` would deconstruct `Exe::WasmBindgen(project_root)` (or `Exe::WasmBindgen { project_root }`), create a `CommandWasmBindgen(project_root)` (or `CommandWasmBindgen { project_root }`) and call `Command::exe_meta` on that, without modifying the trait.

Since the current API treats all `Exe`s and `Command`s the same, I wasn't sure which diverging solution you would prefer.
I personally prefer the second option, for it is cleaner and more aesthetic. But the first approach might be preferable in case the versions of `dart-sass`, `tailwindcss` and `wasm-opt` might in the future also be determined by inspecting the project, which would make passing the project root path to all command resolvers useful.

Please let me know if I should modify the MR.

**Edit: I provided the alternate approach (with I prefer) as https://github.com/leptos-rs/cargo-leptos/pull/590.**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -584,6 +584,7 @@ dependencies = [
  "camino",
  "cargo-config2",
  "cargo-generate",
+ "cargo-lock",
  "cargo_metadata",
  "clap",
  "clap_complete",
@@ -622,6 +623,18 @@ dependencies = [
  "wasm_split_cli_support",
  "which",
  "zip",
+]
+
+[[package]]
+name = "cargo-lock"
+version = "11.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cf53e0ebbbc6e45357b199f3b213f3eb330792c8b370e548499f5685470ecb11"
+dependencies = [
+ "semver",
+ "serde",
+ "toml 0.9.8",
+ "url",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,7 @@ axum = { version = "0.8.4", features = ["ws"] }
 notify-debouncer-full = "0.5.0"
 which = "8.0"
 cargo_metadata = { version = "0.21.0", features = ["builder"] }
+cargo-lock = "11"
 serde_json = "1.0"
 reqwest = { version = "0.12.22", features = [
   "blocking",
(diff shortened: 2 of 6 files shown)


LeoniePhiline:
Please stand by, I'm preparing the other approach as separate PR.

Edit: the alternate approach (which I prefer) can be reviewed at https://github.com/leptos-rs/cargo-leptos/pull/590.

Issue #591, opened 2025-11-25
Missing Android Support


podsvirov:
I added the [cargo-leptos](https://github.com/termux/termux-packages/tree/master/packages/cargo-leptos) package to the [Termux](https://termux.dev/) distribution. And with a small patch, it worked until the update to version `0.3.0`.

What about adding support for `Android` devices?

gbj:
I know almost nothing about this topic (running ordinary binaries on Android).

The only change included in 0.3.0 was a change to stop compiling `wasm-bindgen-cli-support` into the binary, and instead to use the external `wasm-bindgen-cli` tool. Is the problem that there is an issue installing that onto Android automatically? If so, can it be installed with `cargo install` or with its own binary from `wasm-bindgen`?

podsvirov:
> Is the problem that there is an issue installing that onto Android automatically?

Yes! (in my case: target_os=android, target_arch=aarch64)

> If so, can it be installed with `cargo install`

Yes!

I have `wasm-bindgen` executable in `$HOME/.cargo/bin/` directory.

gbj:
A PR to support Android with the necessary changes would be welcome. As I said, I do not know anything about the topic of Android CLI tools.

podsvirov:
> As I said, I do not know anything about the topic of Android CLI tools

It's easy. The Termux project give back Linux on Android.

> A PR to support Android with the necessary changes would be welcome

I have created [PR](https://github.com/termux/termux-packages/pull/27430) that requires manual install `wasm-bindgen` on `Android`.

_NOTE_: All patches will collected in the [cargo-leptos](https://github.com/termux/termux-packages/tree/master/packages/cargo-leptos) package directory. Then I will create `PR` to this/upstream repo.

Issue #560, opened 2025-08-19
Fails at the runtime on FreeBSD: unsupported OS


yurivict:
```
xyz@xyz:/rust/leptos/crud $ /usr/local/bin/cargo-leptos build
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.29s
       Cargo finished cargo build --package=frontend --lib --target-dir=/rust/leptos/crud/target/front --target=wasm32-unknown-unknown --no-default-features
       Front generating JS/WASM with wasm-bindgen
    Finished generating JS/WASM for front in 718.727884ms
The application panicked (crashed).
Message:  called `Result::unwrap()` on an `Err` value:
   0: unsupported OS

Location:
   src/ext/util.rs:14
```

Version: 0.2.42
FreeBSD 14.3

benwis:
This is correct, we've never supported FreeBSD or BSD in general, and noone's asked for it till now. Happy to accept a PR to add in support for it though.

ccmvn:
I just ran into the same issue on FreeBSD 13.5. It would be unfortunate if FreeBSD weren’t supported, especially since many of us use it in production environments.

<img width="742" height="674" alt="Image" src="https://github.com/user-attachments/assets/4301e788-fca3-4c60-85ae-48115502ecfa" />

gbj:
PR #595 may help resolve this issue, by refactoring a bit of the binary-download code. As long as you have `wasm-bindgen-cli` installed it should no longer cause the runtime issue.

PR #593, opened 2025-11-27
fix: android support

podsvirov:
Closes: #591

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,8 @@ install it with the `no_downloads` feature enabled to prevent cargo-leptos from 
 
 > `cargo install --features no_downloads --locked cargo-leptos`
 
+On Android OS cargo-leptos never trying to download any dependencies and you should do it himself.
+
 <br/>
 
 ## Completions
--- a/src/ext/exe.rs
+++ b/src/ext/exe.rs
@@ -222,8 +222,12 @@ impl Exe {
 
         let path = if let Some(path) = meta.from_global_path() {
             path
-        } else if cfg!(feature = "no_downloads") {
-            bail!("{} is required but was not found. Please install it using your OS's tool of choice", &meta.name);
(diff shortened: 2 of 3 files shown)


gbj:
I'm not sure this is entirely the right approach to fix the problem you're having. You say you have `wasm-bindgen` installed, but it is not being found with [this check,](https://github.com/leptos-rs/cargo-leptos/blob/main/src/ext/exe.rs#L223-L225) so it's falling back to the other branches

```rust
        let path = if let Some(path) = meta.from_global_path() {
            path
        } else if cfg!(feature = "no_downloads") {
```
This is just [calling `which::which()`](https://github.com/leptos-rs/cargo-leptos/blob/main/src/ext/exe.rs#L38-L40)

If that's not able to find the binary you have installed, perhaps we should add another branch that checks in another appropriate location to find the binary?

podsvirov:
> ou say you have `wasm-bindgen` installed, but it is not being found with [this check](https://github.com/leptos-rs/cargo-leptos/blob/main/src/ext/exe.rs#L223-L225)

I have a completely different experience. In my case, `wasm-bindgen` was found.

But it may not come to that because [bail!](https://github.com/leptos-rs/cargo-leptos/blob/main/src/ext/exe.rs#L617-L622).

podsvirov:
With these changes I can successfully build and run leptos-based application from [starter template](https://github.com/leptos-rs/start-axum-workspace) on Android device in [Termux](https://termux.dev/) environment.

gbj:
Ah okay, thanks for your patience. I did not write this part of the code base and have not touched it other than adding wasm-bindgen support recently.

Could you test this PR for me as an alternative approach? #595 This should refactor to allow us to support *any* platform where `which` can find the binary, even if downloads for that target are not supported. I think this is more in line with the intent for this feature, and it avoids having to special-case Android support. It may also address #560, I'm not sure.

podsvirov:
Done in #595.

Issue #587, opened 2025-11-21
APFS event coalescing causes Leptos asset copy loop on macOS


oxideNeko233:
## Description

Currently, Leptos uses `fs::copy` to clone asset files into the dist directory when they are updated.

However, on APFS, the FSEvents API may merge multiple recent events into a single emitted event. As a result, after `fs::copy` is called, APFS may emit an additional "modify" event, event that does not correspond to a real file modification..

## Observed Behavior (from `notify` crate logs)

Run `echo "1" > a.css`, `notify` produce following logs:

```
[notify] FSEvent: path = `/path/to/leptos-ssr/public/a.css`, flag = StreamFlags(INODE_META_MOD | ITEM_MODIFIED | IS_FILE)
```

If you wait long enough (~5s) and then run `cp -c a.css ../target/site/pkg`, which equivalent to fs::copy, you get only the clone event:

```
[notify] FSEvent: path = `/path/to/leptos-ssr/public/a.css`, flag = StreamFlags(IS_FILE | ITEM_CLONED)
```

But if you run the clone shortly after the previous write, APFS emits a *merged* event:

```
[notify] FSEvent: path = `/path/to/leptos-ssr/public/a.css`, flag = StreamFlags(INODE_META_MOD | ITEM_MODIFIED | IS_FILE | ITEM_CLONED)
```

In this case, even though we only cloned the file, APFS still sends an extra `ITEM_MODIFIED`  event due to event coalescing.

## Impact on Leptos

This leads to a looped behavior:

1. Leptos detects a change
2. Leptos clones the asset file with fs::copy
3. APFS emits a merged "modify" event for the copied file
4. Leptos interprets this as another asset update
5. Leptos clones the file again
6. APFS emits another "modify" event
7. …and so on

This results in an infinite copy-trigger loop.

## Related Issues

- https://github.com/notify-rs/notify/issues/272
- https://github.com/notify-rs/notify/issues/715

PeteX:
I should have included more of an explanation, sorry. I'm also on MacOS and I'm affected by the same issue. I eventually managed to come up with a work-around, which is in the pull request I posted yesterday. (I really don't know why MacOS generates these duplicate notifications—cloning a file should be analogous to reading it.)

Please let me know what you think of the PR.

PR #600, opened 2025-12-07
Work around repeated watch notifications on MacOS (#587)

PeteX:
As discussed in #587, MacOS sends file modification events when files are cloned from the asset directory to the built site. The spurious events are difficult to distinguish from the genuine ones, so rather than attempt to do that, I record the file modification time when an event is received. If a second event is received, but the file mtime has not changed, it can be assumed to be spurious and discarded.

Please let me know what you think—is this the approach you'd like to take to resolving this issue?

### DIFF
--- a/src/service/notify.rs
+++ b/src/service/notify.rs
@@ -103,8 +103,7 @@ fn handle(
                 .iter()
                 .filter_map(|p| {
                     let mtime = fs::metadata(&p)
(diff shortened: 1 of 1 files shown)


oxideNeko233 (approved):
LGTM 👍

PeteX:
Thank you! 👍

Issue #599, opened 2025-12-07
cargo-leptos serve error message on css parsing: Unexpected token Function("--spacing") at :2304:21


qknight:
# improve the error message

```bash
cargo leptos serve
...
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 15.04s
       Cargo finished cargo build --package=frontend --lib --target-dir=/home/nixos/planicus/target/front --target=wasm32-unknown-unknown --no-default-features
       Front generating JS/WASM with wasm-bindgen
    Finished generating JS/WASM for front in 2.959265715s
    Tailwind finished tailwindcss --input style/style.css --config tailwind.config.js --output /home/nixos/planicus/target/tmp/tailwind.css
Error: 
   0: at `/build/source/src/command/serve.rs:8:46`
   1: Unexpected token Function("--spacing") at :2304:21

Location:
   /build/source/src/compile/style.rs:96

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
```

# code generated

the generated style might be broken or some weird edge-case css. chatgpt claims it is valid css but i'm not sure yet.

`tailwindcss --input style/style.css --config tailwind.config.js --output /home/nixos/planicus/target/tmp/tailwind.css`

the code on line 2304 is:

```javascript
.\[--anchor-gap\:--spacing\(3\)\] {
  --anchor-gap: var(--spacing(3));
}

.\[appearance\:textfield\] {
  -webkit-appearance: textfield;
     -moz-appearance: textfield;
          appearance: textfield;
}
```

# wish

We should have a better error message, so:

```
Error: 
   0: at `/build/source/src/command/serve.rs:8:46`
   1: Unexpected token Function("--spacing") at :2304:21
```

should become

```
Error: 
   0: at `/build/source/src/command/serve.rs:8:46`
   1: Unexpected token Function("--spacing") at /home/nixos/planicus/target/tmp/tailwind.css:2304:21
       .\[--anchor-gap\:--spacing\(3\)\] **{**
```

qknight:
```rust
  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
                                ⋮ 2 frames hidden ⋮                               
[...]

PR #602, opened 2025-12-08
#599 Use CSS even if parser failed to parse it

qknight:
# Motivation

Because lightningcss fails to parse the tailwindcss  v3.4.17 output, I can't use `cargo leptos serve`.

The proposed solution in this PR is to use the 'raw' input CSS without parsing or minification as a fallback.

`cargo leptos serve -vv` shows this now:

image

https://github.com/leptos-rs/cargo-leptos/issues/599

This fallback concept is also found in **trunk**. A huge difference: **trunk does not minify the tailwindcss**, as attempted here in `cargo leptos`. In trunk it calls the **tailwindcss with -m** on release builds, which minimizes it already. Thus in trunk, it uses the CSS artifact without any parsing, I checked it, it is the same CSS code and same parser `lightningcss` in both tools.

# cargo-leptos

    cargo leptos serve --release
    warning: `frontend` (lib) generated 22 warnings
    error: could not compile `frontend` (lib) due to 1 previous error; 22 warnings emitted
        Tailwind finished tailwindcss --input style/style.css --config tailwind.config.js --output /home/nixos/planicus/target/tmp/tailwind.css --minify

So we could be faster when we don't minify an already minified target.

/home/nixos/cargo-leptos/src/compile/tailwind.rs

```rust
pub async fn tailwind_process(
    proj: &Project,
    cmd: &str,
    tw_conf: &TailwindConfig,
) -> Result<(String, Command)> {
    let tailwind = Exe::Tailwind.get().await.dot()?;

    let mut args = vec!["--input", tw_conf.input_file.as_str()];

    if let Some(config_file) = tw_conf.config_file.as_ref() {
        args.push("--config");
        args.push(config_file.as_str());
    }

    args.push("--output");
    args.push(tw_conf.tmp_file.as_str());

    if proj.release {
        // minify & optimize
        args.push("--minify");
    }

    let line = format!("{} {}", cmd, args.join(" "));
    let mut command = Command::new(tailwind);
    command.args(args);

    Ok((line, command))
}
```

# trunk

Here is what trunk does:

## the call to minify_css(bytes),

https://github.com/trunk-rs/trunk/blob/f1ee3d4032cb939b8513d1d8fabfcb84ce46d811/src/pipelines/mod.rs#L312

## the minify_css implementation

https://github.com/trunk-rs/trunk/blob/f1ee3d4032cb939b8513d1d8fabfcb84ce46d811/src/processing/minify.rs#L18C8-L18C18

### DIFF
--- a/src/compile/style.rs
+++ b/src/compile/style.rs
@@ -92,26 +92,42 @@ async fn process_css(proj: &Project, css: String) -> Result<Product> {
     let browsers = browser_lists(&proj.style.browserquery).wrap_err("leptos.style.browserquery")?;
     let targets = Targets::from(browsers);
 
-    let mut stylesheet =
-        StyleSheet::parse(&css, ParserOptions::default()).map_err(|e| eyre!("{e}"))?;
+    let filename: String = if let Some(tw) = proj.style.tailwind.clone() {
+        tw.tmp_file.to_string()
+    } else {
+        proj.style.file.as_ref()
+            .map(|f| f.source.to_string())
+            .unwrap_or_default()
+    };
 
-    if proj.release {
-        let minify_options = MinifyOptions {
-            targets,
-            ..Default::default()
-        };
-        stylesheet.minify(minify_options)?;
-    }
-
-    let options = PrinterOptions::<'_> {
-        targets,
-        minify: proj.release,
+    let parse_options = ParserOptions {
+        filename,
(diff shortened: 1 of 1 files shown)


gbj (approved):
This looks good to me, overall! Thanks for the PR.

PR #606, opened 2025-12-21
fix: io-errors from .gitignore ignored

### LINKED ISSUE #605: Missing `.gitignore` silently causes `cargo leptos watch` to hang
If you don't have a `.gitignore` then `cargo leptos watch` will hang and ignore ctrl+c. I assume it's picking up changes (probably recursively) in target.

It took me a while to figure out #604 so I would guess other people might run into this issue as well. Even just a warning that 'the number of files being watched is excessive' would have been enough to clue my grug-brain in. I only figured it out while creating a repro commit and realizing I was missing a `.gitignore`. The output still happily printed `Creating ignore list from '.gitignore' file` despite the rather important lack of that file's existence.

A slightly more refined version might have a default ignore set it reverts back to if a `.gitignore` is missing (things like Mercurial do in fact exist lol) and a warning to that effect.

benedikt-schaber:
Fixes #605
## Intended behavior
When running `cargo-leptos watch` the following error should be printed if the `.gitignore` file is missing:
```
Failed reading '.gitignore' file in the working directory: {err}
This causes the watcher to work expensively on file changes like changes in the 'target' path.
Create a '.gitignore' file and exclude common build and cache paths like 'target'
```

## Current behavior
IO-errors are ignored, the error is not printed if the file is missing.

## Fix
From `Gitignore::new`:
```
    /// Note that I/O errors are ignored. For more granular control over
    /// errors, use `GitignoreBuilder`.
```
Now it is using `GitignoreBuilder` and does not ignore `IO` errors, further in case the file is there but invalid I added a more specific error message, otherwise the behavior remains the same (I pretty much just inlined `Gitignore::new`).

## Further work
Now that the error is working this means, that it will also be printed if running `cargo-leptos` with the current working directory in one of the examples. Should we perhaps also try to look in a parent folder for a `.gitignore` file or something similar? I could also create a PR for that.

### DIFF
--- a/src/service/notify.rs
+++ b/src/service/notify.rs
@@ -294,7 +294,14 @@ impl GitAwareWatcher {
     fn new_gitignore(gitignore_path: &Path) -> Gitignore {
         info!("Creating ignore list from '.gitignore' file");
 
-        let mut builder = GitignoreBuilder::new(gitignore_path.parent().unwrap_or(Path::new("/")));
+        let glob_dir = match gitignore_path.parent() {
+            Some(dir) => dir,
+            None => {
+                error!("Could not determine parent directory of '.gitignore' file\nThis causes the watcher to work expensively on file changes like changes in the 'target' path.\nCreate a '.gitignore' file and exclude common build and cache paths like 'target'");
+                return Gitignore::empty();
+            }
+        };
+        let mut builder = GitignoreBuilder::new(glob_dir);
         let err = builder.add(gitignore_path);
         if let Some(err) = err {
             error!("Failed reading '.gitignore' file in the working directory: {err}\nThis causes the watcher to work expensively on file changes like changes in the 'target' path.\nCreate a '.gitignore' file and exclude common build and cache paths like 'target'");


src/service/notify.rs
```diff
@@ -294,13 +294,18 @@ impl GitAwareWatcher {
     fn new_gitignore(gitignore_path: &Path) -> Gitignore {
         info!("Creating ignore list from '.gitignore' file");
 
-        let (gi, err) = Gitignore::new(gitignore_path);
-
+        let mut builder = GitignoreBuilder::new(gitignore_path.parent().unwrap_or(Path::new("/")));
```
sabify: Why are you trying to look up the parent directory of the .gitignore path, and even weirder, why are you adding the filesystem root path?

benedikt-schaber: I tried to keep the behavior as-is as far as possible. This is just copied from `Gitignore::new`, i.e. this did not change, but was just inlined, because we cannot use `Gitignore::new` if we do not want to ignore IO-Errors.

So, I can give an explanation if you want, but don't know if it is necessary, since it did not change.

src/service/notify.rs
```diff
@@ -294,13 +294,18 @@ impl GitAwareWatcher {
...
         }
-
-        gi
+        match builder.build() {
+            Ok(gi) => gi,
+            Err(err) => {
+                error!("Failed building gitignore matcher: {err}\nThis causes the watcher to work expensively on file changes like changes in the 'target' path.\nCreate a '.gitignore' file and exclude common build and cache paths like 'target'");
+                Gitignore::empty()
+            }
+        }
```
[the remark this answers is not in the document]
benedikt-schaber: An error can either occur, when the path is added (file missing), this is handled above. Or when the file is there but e.g. corrupted. Errors like this are handled here.
If you find the second part of the error message confusing I can adapt that.

src/service/notify.rs
sabify: Ah, I just got it. 

Please just do not fallback to the filesystem's root path. Simply handle the error by printing another message. Something like:

```rust
let dir = match gitignore_path.parent() {
    Ok(dir) = dir,
    Err(err) = {
        // print error
        return Gitignore::empty();
    }
};
let mut builder = GitignoreBuilder::new(dir));
```

src/service/notify.rs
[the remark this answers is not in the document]
sabify: You are right. I just missed the builder pattern...

src/service/notify.rs
benedikt-schaber: Updated. I found it a bit difficult to get into the project's style of error writing and a bit difficult to imagine a user running into this problem, so I would be glad on feedback if the error message sounds a bit off.

PR #611, opened 2025-12-23
fix: respect site_pkg_dir (closes #609, closes #495)

### LINKED ISSUE #495: `site-pkg-dir` conflict with `assets-dir` if `site-pkg-dir` contains `/`
When a project set both `site-pkg-dir` and `asset-dir` like this one:

```
[[workspace.metadata.leptos]]
site-pkg-dir = "static/pkg"  # set sub folder in here 
assets-dir = "assets"        # may contains some thing like `favicon.ico` and `static/my_image.png` and should not conflicted
```

The generated wasm & js will be removed due to following code:

https://github.com/leptos-rs/cargo-leptos/blob/9bab0d7708f53b53592fdb49133ab79f8d1858c5/src/compile/assets.rs#L67-L74

https://github.com/leptos-rs/cargo-leptos/blob/9bab0d7708f53b53592fdb49133ab79f8d1858c5/src/compile/assets.rs#L81-L87

### LINKED ISSUE #609: `site_pkg_dir` doesn't seem to allow a sub-directory.
**Describe the bug**

[_Running on an Intel Mac_] If the `site_pkg_dir` config option is set to a value with a sub-directory, only the css is built.

**Leptos Dependencies**

```toml
leptos = { version = "0.8.2" }
leptos_meta = { version = "0.8.2" }
leptos_actix = { version = "0.8.2", optional = true }
leptos_router = { version = "0.8.2" }
```

**To Reproduce**

1. Use the actix starter: `cargo leptos new --git https://github.com/leptos-rs/start-actix`
2. Create a `.env` file and set the `LEPTOS_SITE_PKG_DIR` var to a value with a sub-directory (e.g. `_app/pkg`)
3. Run application with `cargo-leptos` (`cargo-leptos serve`)
4. Check in `target/site/_app/pkg` and see that the `.js`, `.d.ts`, and `.wasm` files are missing

**Screenshots**

Not applicable

**Next Steps**

- [ ] I will make a PR
- [x] I would like to make a PR, but need help getting started
- [ ] I want someone else to take the time to fix this
- [x] This is a low priority for me and is just shared for your information

**Additional context**

I'm happy to look into this as I'm guessing it's a simple issue with parsing the file path or with ensuring that the sub-directory exists before trying to write the js/wasm files (maybe here? `leptos/src/hydration/mod.rs`). But also happy if someone else wants to look into it. It's really not a pressing issue for me.

sabify:
Closes #609
Closes #495

### DIFF
--- a/src/ext/fs.rs
+++ b/src/ext/fs.rs
@@ -127,7 +127,7 @@ async fn cp_dir_all(src: impl AsRef<Utf8Path>, dst: impl AsRef<Path>) -> Result<
             let to = from.rebase(src, &dst)?;
 
             if entry.file_type()?.is_dir() {
-                self::create_dir(&to).await?;
+                self::create_dir_all(&to).await?;
                 dirs.push_back(from);
             } else {
                 self::copy(from, to).await?;


benedikt-schaber:
I also took a quick look at that issue. Probably an intentional design decision from you, but just wanted to point out, that this solution currently would not  do a "deep clean" of the directory. I.e. if the `pkg_dir` is `app/pkg` it wouldn't clean `app/example`.

Further, the `reserved` code in that file is currently broken as well. I thought about opening a PR on that too. But I could image that you would just want to fix this in this PR as well? Tell me if not, then I will quickly spin up a separate PR.

sabify:
AFAIK, the `site-root` will be removed and reconstructed on each iteration. So, there won't be any subdirectory other than `site-pkg-dir` that will be preserved by this PR.

benedikt-schaber:
Ah, that's right of course, done in build_proj - my bad :)

gbj (approved):
Thanks!

PR #615, opened 2025-12-29
feat: warn on unknown fields in `metadata.leptos`

### LINKED ISSUE #266: Warn about invalid `package.metadata.leptos` parameters
To prevent things like https://github.com/leptos-rs/start/pull/43

benedikt-schaber:
Fixes #266

Also removed unrecognized field in example's `Cargo.toml` and some unnecessary clones, because `serde_json::from_value` was used instead of the fields from the `Deserialize` trait.

`serde-ignored` would be an alternative approach to the current implementation, however, that doesn't support `serde(flatten)` currently and support for that is blocked by a Rust RFC, so that will likely take some time.

### DIFF
--- a/src/config/project.rs
+++ b/src/config/project.rs
@@ -369,30 +369,30 @@ impl ProjectConfig {
         cargo_metadata: &Metadata,
     ) -> Result<Self> {
         let mut conf: ProjectConfig = serde_json::from_value(metadata.clone())?;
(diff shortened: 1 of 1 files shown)


benedikt-schaber:
@gbj Sorry for the small delay, updated now.

gbj (approved):

gbj:
Thanks!

PR #619, opened 2025-12-30
fix: end2end-dir not based on config_dir

### LINKED ISSUE #386: end2end-dir is not relative to current directory
I'm using `cargo leptos` based off of the project rather than workspace setup, as the project is deeper within an existing workspace.

When running `cargo leptos end-to-end`, with the following defined in the `Cargo.toml`, it attempts to launch tests relative to the parent Workspace, rather than relative to the current directory:
```toml
end2end-cmd = "pytest -c end2end/pytest.ini"
end2end-dir = "."
```

Error:
`FileNotFoundError: [Errno 2] No such file or directory: '/workspaces/src/<parent-workspace>/end2end/pytest.ini'`

The actual location it should be running is `/workspaces/src/<parent-workspace>/code/rust/ui/<my-project>`

benedikt-schaber:
Fixes #386

### DIFF
--- a/src/config/end2end.rs
+++ b/src/config/end2end.rs
@@ -13,11 +13,11 @@ impl End2EndConfig {
     pub fn resolve(config: &ProjectConfig) -> Option<Self> {
         let cmd = &config.end2end_cmd.to_owned()?;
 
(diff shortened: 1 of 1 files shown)


gbj (approved):
Looks good, thanks.

PR #623, opened 2026-01-02
fix: Apply lightningcss stylesheet minification in both debug and release

### LINKED ISSUE #622: Invalid usage of lightningcss in debug mode breaks variables referencing light-dark
## Problem description

As mentioned in this comment: https://github.com/parcel-bundler/lightningcss/issues/935#issuecomment-2739325537

And this fix: https://github.com/vercel/next.js/pull/77313

`StyleSheet::minify` needs to be called every time prior to calling `StyleSheet::to_css`. This is because `minify` does a lot more than just minifying (removing whitespace). It also adds things like variable declarations in the appropriate scope for certain functions if the targets list is set to a certain value which triggers compatibility options being applied. In the case of `light-dark`, compiling the following CSS using cargo-leptos behaves as follows:

Input:
```css
.foo {
	color-scheme: light dark;
}
```

Expected (as found in a [lightningcss unit test](https://github.com/parcel-bundler/lightningcss/blob/400f705e63e139c326f480aed11e1416f5a3a61f/src/lib.rs#L29732)):
```css
.foo {
    --lightningcss-light: initial;
    --lightningcss-dark: ;
    color-scheme: light dark;
}

@media (prefers-color-scheme: dark) {
    .foo {
        --lightningcss-light: ;
        --lightningcss-dark: initial;
    }
}
```

Actual (output in pkg css after running `cargo leptos watch`):
```css
.foo {
  color-scheme: light dark;
}
```
Running in release mode with `cargo leptos watch --release` produces the following:

```css
.foo{--lightningcss-light:initial;--lightningcss-dark: ;color-scheme:light dark}@media (prefers-color-scheme:dark){.foo{--lightningcss-light: ;--lightningcss-dark:initial}}
```

Which is both transformed correctly and minified, as expected.

---

## Cause

The offending code can be found in [compile::style.rs#L110](https://github.com/leptos-rs/cargo-leptos/blob/7ccf2a1276d1f6ea97bf4146f9d43a54824ad767/src/compile/style.rs#L110) where the poorly named `minify` is being called conditionally only if release mode is enabled, after which `to_css` is called:

```rust
...
if proj.release {
                let minify_options = MinifyOptions {
                    targets,
                    ..Default::default()
                };
                stylesheet.minify(minify_options)?;
            }
            let options = PrinterOptions::<'_> {
                targets,
                minify: proj.release,
                ..Default::default()
            };
            stylesheet.to_css(options)?.code
...
```

---

## Solution

A short-term solution is to simply remove the release mode check and call `minify` every time, which is the route others have taken. However, ideally the lightningcss code should not allow this behavior to occur at all, as it's quite likely someone might come along and make the same mistake in the future. I am opening an issue in the lightningcss repo for potential improvements which I will attach below, and will open a PR for the short-term fix.

dVeon-loch:
- Fixes #622
- See https://github.com/parcel-bundler/lightningcss/issues/1118 for more information

### DIFF
--- a/src/compile/style.rs
+++ b/src/compile/style.rs
@@ -107,13 +107,11 @@ async fn process_css(proj: &Project, css: String) -> Result<Product> {
 
     let css: String = match StyleSheet::parse(&css, parse_options) {
         Ok(mut stylesheet) => {
-            if proj.release {
-                let minify_options = MinifyOptions {
+            stylesheet.minify(MinifyOptions {
                     targets,
                     ..Default::default()
-                };
-                stylesheet.minify(minify_options)?;
-            }
+            })?;
+
             let options = PrinterOptions::<'_> {
                 targets,
                 minify: proj.release,


gbj (approved):
Thanks for discovering this, and for making the PR. This seems like a fine approach to me.

Issue #629, opened 2026-02-23
Create a release for version 0.3.4


CodingFlow:
Should an official release on the GitHub repository be created for version 0.3.4? It might confuse some mirrors not to have one and it would be nice to have the direct download option for binaries.

PR #630, opened 2026-02-24
fix(CI): cargo-dist

sabify:
@gbj, you need to cut a new release to take this into effect.

The process of updating the cargo-dist would be only `cargo dist init --yes` after this PR and it will automatically update both `release.yml` and `dist-workspace.toml`.

Closes: #629


gbj:
Thanks @sabify, can confirm this fixed it for a new 0.3.5 release. I really appreciate all your help.

PR #631, opened 2026-02-24
chore: upgrade deps

sabify:


sabify:
The failed `Release` workflow has been fixed in #630.

Issue #641, opened 2026-04-12
`wasm_bindgen` module imports are not hashed inside `.wasm` binary when `LEPTOS_HASH_FILES=true`


johomo:
First off, thanks for all the work on this great framework. As a Rust programmer, it's a real joy to build for the web using Rust and Leptos.

Please let me know if this is actually a bug with cargo-leptos, or if I should move this issue to a different Leptos repository.

## Issue description

When building a Leptos SSR project with `LEPTOS_HASH_FILES=true`, `cargo-leptos` correctly hashes the filenames of the generated assets (JS, WASM, CSS). However, if the project uses `wasm_bindgen` to import a local JavaScript file via `#[wasm_bindgen(module = "/js/javascript.js")]`, the resulting `.wasm` binary still contains an internal import reference to the unhashed filename.

This causes a runtime error in the browser because it tries to fetch the original filename (which doesn't exist) instead of the hashed version.

I am available to provide a minimal reproducible example repository if needed.

## Steps to reproduce

1. Create a JavaScript file to be called from Rust:
```javascript
// File ./js/javascript.js
export function javascriptFunction(config) {
    console.log("Hello from the brower");
    // Parameter `config` is not rellevant.
}
```

2. Create a Leptos component that imports this function:
```rust
// Rust component
use leptos::wasm_bindgen::prelude::*;

#[cfg(feature = "hydrate")]
#[wasm_bindgen(module = "/js/javascript.js")]
extern "C" {
    #[wasm_bindgen(js_name = javascriptFunction)]
    fn javascript_function(config: JsValue);
}

// This function cannot be run in server-side (javascript is only available in client-side)
#[cfg(not(feature = "hydrate"))]
fn javascript_function(_: JsValue) {}
```

3. Build the project in release mode with hashing enabled:
```bash
LEPTOS_HASH_FILES=true cargo leptos build --release
```

4. Run the application:
```bash
LEPTOS_HASH_FILES=true cargo leptos serve --release
```

The build completes successfully. The files in `./target/site/pkg` are correctly hashed, including the JS snippet:
```bash
target/site/pkg
|- leptos-ssr-template.4ul9fK4Rkx4nbHf8QDA-lg.wasm
|- leptos-ssr-template_bg.wasm.d.VkYcl6ddV5YtKq5nEt4uNg.ts
|- leptos-ssr-template.5v77LImhofSXRFXocdUK-w.js
[...]

PR #643, opened 2026-04-17
fix: Do not hash files in ./target/site/pkg/snippets/ folder, even with LEPTOS_HASH_FILES=true

johomo:
## Issue description

When building a Leptos SSR project with `LEPTOS_HASH_FILES=true`, `cargo-leptos` correctly hashes the filenames of the generated assets (JS, WASM, CSS). However, if the project uses `wasm_bindgen` to import a local JavaScript file via `#[wasm_bindgen(module = "/js/javascript.js")]`, the resulting `.wasm` binary still contains an internal import reference to the unhashed filename.

This causes a runtime error in the browser because it tries to fetch the original filename (which doesn't exist) instead of the hashed version.

Extended details and discussion here https://github.com/leptos-rs/cargo-leptos/issues/641

## Fix proposed

Do not hash any files in the `./target/site/pkg/snippets/ ` folder.


gbj:
Just to double-check that my understanding of this is correct, because I don't use a lot of `wasm-bindgen` externals like this: previously, we only skipped hashing for *inline* JS snippets by checking for "inline" in the filename; but still (incorrectly) hashed snippets in local JS files because they don't have that in the name? (Edit: And this fixes it by skipping all files that are in snippets)

johomo:
First of all, I'm not an expert in wasm_bindgen nor webassembly. Here's what I found.

* wasm_bindgen generates the `snippets` folder and populates it with all JS files specified in directives such as `#[wasm_bindgen(module = "/js/javascript.js")]` and `#[wasm_bindgen(inline = "...")]`
* wasm_bindgen generates the `wasm` file (as far as I understand), which imports the JS files from the `snippets` folder.
* `cargo-leptos` hashes the JS files in `snippets` folder, but does not update the import statements of the `wasm` files.

Here's the issue: hashing JS files in `snippets` folder break the import statements of the `wasm` file (regardless of whether the file is generated with a `module` argument or an `inline` argument). By not hashing the files in the `snippets` folder, we do not break the imports of the wasm file.

In other words, cargo-leptos was not hashing inline files for the same reason. We are applying this fix now for non-inline JS files too. 

#### A comment about the proposed implementation
As far as I researched, wasm_bindgen populates the `snippets` folder exclusively with the JS files specified by `[wasm_bindgen(module = "/js/javascript.js")]` or `[wasm_bindgen(inline = "...")]`. No other files should be found in this folder (even though I'm not sure if this folder could contain typescript files, and if so, i don't know if typescript files require the same fix). For this reason I propose ignoring the folder completely.

----

Finally, I have amended the PR to correct the username and email of the commit author.

gbj (approved):
Thanks!

gbj:
Thanks. Yes, this looks good to me!

ghost (commented):
@johomo ayudame por favor..

PR #645, opened 2026-04-19
fix(wasm-split): honor configured pkg_dir in split-loader URL

adrianncovaci:


gbj (approved):
Thanks very much! Certainly looks right, and works with default and with custom pkg_dir when I test it.

PR #646, opened 2026-04-30
add [--target] flag to cli build sub command for cross compiling server binary.

kanodecat:
I realize after making this that I could've made a config file for cargo-leptos to read and label the target in variable 
```bash
bin-target-triple = "x86_64-unknown-linux-gnu"
``` 
but its nicer for end users as cli flag for testing or extra options

### DIFF
--- a/src/compile/tests.rs
+++ b/src/compile/tests.rs
@@ -24,6 +24,7 @@ fn release_opts() -> Opts {
         frontend_only: false,
         server_only: false,
         clear: false,
+        target: None,
     }
 }
 fn dev_opts() -> Opts {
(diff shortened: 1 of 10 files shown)


gbj:
I think this already exists as an env var, per the README: `LEPTOS_BIN_TARGET_TRIPLE`.

kanodecat:
yes that's correct and didn't see that until mentioned and works great in bash; I still think adding the '--target' flag is a good idea as it keep with conventions with rust cargo and simplifies user interaction, that's how I expected it to work when I tried cross compiling to arm64 earlier this afternoon. 
image
the flag also gets applied to sub command 'serve' which I'm not sure would serve a purpose given the command is running on the host machine...

PR #647, opened 2026-05-05
Graceful shutdown

lpotthast:
I observed an insufficient shutdown logic after developing an internal dev-tool that required to terminate external resources cleanly before shutting down. Code behind a `ctrl+c` handler would not execute, because cargo-leptos was killing the served app outright instead of giving it a chance to exit on its own.

This PR adds graceful shutdown support for the user-app served by cargo-leptos (enabled by default) with a few new configuration knobs.

I've been sitting on this for a while. Happy to see it merged if it fits.

## New config                                                                                                                                                                
                                                         
CLI flags and `[package.metadata.leptos]` keys (CLI wins as usual):                                                                                                           
                                                         
- `graceful-shutdown` (bool, default `true`) - set `false` to restore the previous immediate-kill behavior.                                            
- `graceful-shutdown-interrupt-timeout-secs` (u64, default `4`) - SIGINT wait timeout on Unix.                                                                            
- `graceful-shutdown-terminate-timeout-secs` (u64, default `4`) - SIGTERM wait timeout on Unix. Reused as the single `CTRL_BREAK_EVENT` wait timeout on Windows.

## Implementation note

Most of the logic lives in the new https://crates.io/crates/tokio-process-tools dependency. It enforces manual termination of a spawned process(group), which we do here, using explicit, configurable timeout budgets. I initially built this library for something different, focused on capturing the processes output as well. The crate was designed around piped IO from the start. Discarding IO completely is now also possible. Should cargo-leptos want to keep the user process inheriting our own IO, an inherit-stdio mode could be upstreamed in tokio-process-tools. But I would argue that we pay very little here for an added bonus of having the structured / programmatically accessible output available for additional processing. This opens up a lot of possibilities for further improvements or integrations.


gbj (approved):
This seems totally reasonable to me. Thanks for the PR!

PR #648, opened 2026-05-09
Graceful shutdown - followup

lpotthast:
Follow-up to PR #647. This simplifies the graceful-shutdown configuration surface to a single timeout and allows users to configure which signal is used to request a graceful shutdown on unix systems.

The previous two-phase Unix shutdown (SIGINT -> SIGTERM, each with its own timeout) didn't make much sense really. That was a misconception on my end. As the default disposition for both signals is "term", the users app would just be killed on the first signal, should it not have registered a handler for it. There is no real possibility to do "escalation" here. We need to decide which signal to send, and send only that one. Sending both signals in sequence would only make sense if we know that both are handled and that it is better to send both than to just send a SIGTERM alone, which should be super rare.

This opens the question: Which is in fact the "correct" signal that cargo-leptos should send to the users application? In this PR, I have kept it at SIGINT. One could argue for that by saying: cargo-leptos is a CLI / dev-tool style application. It is expected that cargo-leptos is terminated by a ctrl+c (it correctly uses `tokio::signal::ctrl_c()`). So "forwarding" that to the application started by cargo-leptos is fine. One could also say that: The started user application is in fact NOT a CLI / dev-tool, but a regular frontend application / webserver serving it. It should probably listen for SIGTERM instead, as that signal would also be issued by orchestrators like Kubernetes or Docker.

But then again: Its probably common that `tokio::signal::ctrl_c()` is used initially to make an app shut down gracefully. But that ONLY registers an interrupt handler (unix), so requires cargo-leptos to send a SIGINT to work, and does not work for Windows at all (tokio-process-tools always spawns the process as a new process group to reliably terminate everything the process might have started as its own children. But that requires sending a CTRL_BREAK_EVENT on Windows, which `tokio::signal::ctrl_c()` also doesn't register a handler for.). Users should probably always handle both signals on unix and run the same graceful logic, just to be safe. I have stated that in the readme.

## Configuration changes

Removed:
  - `graceful-shutdown-interrupt-timeout-secs` / `LEPTOS_GRACEFUL_SHUTDOWN_INTERRUPT_TIMEOUT_SECS`
  - `graceful-shutdown-terminate-timeout-secs` / `LEPTOS_GRACEFUL_SHUTDOWN_TERMINATE_TIMEOUT_SECS`

Added:
  - `graceful-shutdown-timeout-secs` / `LEPTOS_GRACEFUL_SHUTDOWN_TIMEOUT_SECS` (default `10`)
  - `graceful-shutdown-unix-signal` / `LEPTOS_GRACEFUL_SHUTDOWN_UNIX_SIGNAL`
    (`SIGINT` or `SIGTERM`, default `SIGINT`, ignored on Windows)

We also expose all three configs (including `LEPTOS_GRACEFUL_SHUTDOWN`) as env-var options now. I forgot to add that last time.

The new single timeout is `10s`, giving apps a slightly more lenient window before being killed abruptly.

Edit:
I also updated cargo-leptos to register more signals on Unix and Windows. This lets cargo-leptos shut down gracefully more reliably in more environments.

## Dependency bumps

- `tokio-process-tools` 0.10.1 -> 0.11.0


gbj (approved):
This all seems good to me. Thanks for your work!

PR #653, opened 2026-05-29
fix: content-hash the wasm-split loader after rewriting its references

autrilla:
> Disclaimer: this PR was generated with the help of Claude, though I was in the loop and reviewed and changed the direction Claude was taking multiple times.

## Problem

When wasm-splitting is enabled, the generated `__wasm_split.<hash>.js` loader had its filename hash computed from its content **before** its chunk references were rewritten to their hashed names. As a result the hash never reflected the loader's final content, so the loader's filename stayed constant even when chunk contents changed — defeating cache-busting. Browsers could serve a stale loader after a rebuild/redeploy, pointing at chunk hashes that no longer exist.

## Fix

- Exclude the loader from the first hashing pass, rewrite its chunk references, then hash and rename it from the resulting (final) content.
- The main JS is now rewritten **last**, once the loader's final name is known. This removes the previous round-trip where the main JS was rewritten and then patched up again.

## Cleanups included

- Extracted the loader's placeholder filename into a shared `WASM_SPLIT_LOADER_PLACEHOLDER` constant, so the producer (`wasm_split_tools`) and consumer (`compile::hash`) reference one source of truth instead of duplicating the literal.
- Extracted a `content_hash` helper to de-duplicate the base64url-unpadded MD5 logic that previously appeared in several places.

## Testing

- Added a regression test (`wasm_split_loader_filename_reflects_post_rewrite_content`) asserting that two builds with identical loader templates but different chunk contents produce different loader filenames — which fails on the old code.
- Verified end-to-end on a real split-enabled Leptos project: the loader filename's hash now matches the MD5 of its actual bytes, all `__wasm_split` references resolve, and changing a split chunk's content changes the loader filename.

### DIFF
--- a/src/compile/hash.rs
+++ b/src/compile/hash.rs
@@ -1,6 +1,6 @@
 use crate::{
     config::Project,
-    ext::{eyre::CustomWrapErr, PathBufExt},
+    ext::eyre::CustomWrapErr,
     internal_prelude::*,
 };
 use base64ct::{Base64UrlUnpadded, Encoding};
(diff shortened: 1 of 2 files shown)


gbj:
Thanks for the PR. I didn't see this until I'd merged the same thing in 9f13fe7c76ef023f17b84c2122fafae761373739

PR #654, opened 2026-06-07
Add frontend Cargo target directory override

steveej:
Changes:
- Adds a `front-target-dir` metadata option for selecting the Cargo target directory used by frontend WASM builds.
- Resolves relative `front-target-dir` values from the workspace root and supports existing Cargo target directory markers.
- Uses the selected frontend target directory for both the Cargo `--target-dir` argument and hydrate WASM artifact lookup.
- Keeps the unset-option default at `<resolved Cargo target dir>/front` and documents the override.
- Adds unit coverage for default, marker-based, and workspace-relative frontend target directory resolution.

Why:
- This lets workflows opt into placing and reading WASM artifacts from the resolved Cargo target root when the dedicated frontend subdirectory does not match their build-cache layout.

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,6 +220,13 @@ lib-profile-dev = "my-debug-profile"
 # Optional. Defaults to false prior to 0.2.3, unconditionally enabled (with the setting becoming deprecated) since 0.2.3 and #216
 separate-front-target-dir = true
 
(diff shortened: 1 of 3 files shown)


gbj (approved):
Looks good, thanks!

Issue #659, opened 2026-06-25
wasm_split broken when non-standard pkg-dir is used


Ofenhed:
Building a split project with a non-standard `LEPTOS_SITE_PKG_DIR` breaks, as `__wasm_split._[...]_.js` tries to load an invalid path. I think the problem originates in [wasm_split_tools/mod.rs:16](https://github.com/leptos-rs/cargo-leptos/blob/e2099d97fb6ed39f1bdd24218b7885438cf6c704/src/wasm_split_tools/mod.rs#L16), where it loads the main javascript file without checking the environment equivalents.

PR #660, opened 2026-06-27
feat: decouple public pkg URL from on-disk pkg dir

sabify:
Paired with: https://github.com/leptos-rs/leptos/pull/4817

Add a `site-pkg-url` config (env `LEPTOS_SITE_PKG_URL`, default `pkg`) as
the public URL path under which the pkg assets are served, separate from
`site-pkg-dir` (the on-disk location). An absolute `site-pkg-dir` is a
server-side path and is no longer exposed in client URLs.

- Add `ProjectConfig::site_pkg_url`, overlay it from the env var, and
  export `LEPTOS_SITE_PKG_URL` to the running app.
- Add `Site::pkg_url` and `Site::pkg_url_path()`, mirroring
  `LeptosOptions::pkg_url_path` in leptos_config.
- Build the wasm-split main-module URL from `pkg_url_path()` so the split
  loader never re-imports via an absolute on-disk path.
- Update config snapshots and env tests for the new variable.


gbj:
It's not clear to me how this relates to #659, which can be fixed more simply (see #666)

sabify:
I was thinking about this feature for a long time to decouple the serving path from package directory, and I thought that issue could be part of this feature. IMO, It resolves that, but you provided a hotfix for that. I just edited the description and removed "Closes: ...".

PR #662, opened 2026-07-02
chore: bump wasm_split_cli_support to 0.2.2 (retry failed split-chunk loads)

JohnHarrison:
## What

Bumps `wasm_split_cli_support` `0.2.1` → `0.2.2` (Cargo.toml floor + lockfile).

## Why

0.2.1 generates a `__wasm_split.js` whose loader memoizes **rejected** load promises: a single transient network failure while fetching a split chunk (flaky connection, brief offline, server blip) permanently disables that chunk for the session — every later attempt to use the lazy-loaded code re-throws the cached rejection, and only a full page reload recovers. In a `--split` Leptos app this bricks lazy routes after one bad fetch.

0.2.2 memoizes successes only, so the next invocation after a failure retries the fetch ([WorldSEnder/wasm-split-prototype#36](https://github.com/WorldSEnder/wasm-split-prototype/pull/36), released 2026-06-28). This is the loader-side half of the recovery story for leptos-rs/leptos#4785 (surfacing chunk-load failures to `ErrorBoundary`), but it stands on its own: it turns a permanent brick into a plain retry regardless of how the Rust side handles the error.

The lockfile bump is the part that reaches users: `Cargo.toml`'s caret req already permits 0.2.2, but `cargo install cargo-leptos --locked` (the documented install) pins 0.2.1 until the lockfile moves.

## Compatibility

- `cargo check` / `cargo build` pass; the fixed loader template is confirmed embedded in the built binary.
- 0.2.2 also adds an optional link-time canary. Verified it degrades gracefully: the canary import is only emitted when the app's wasm was built with `wasm_split_helpers` ≥ 0.2.2 in debug; wasm built against older helpers carries no feature subsection and splits exactly as before, so this is a drop-in for existing apps.
- Transitive updates limited to what 0.2.2 requires: `wasm-encoder`/`wasmparser` 0.248 → 0.252.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01R5Eb5TeDybfYmnuHeHPnhF

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6281,12 +6281,12 @@ dependencies = [
 
 [[package]]
 name = "wasm-encoder"
-version = "0.248.0"
+version = "0.252.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ac92cf547bc18d27ecc521015c08c353b4f18b84ab388bb6d1b6b682c620d9b6"
+checksum = "8185ae345fa5687c054626ff9a50e7089797a343d9904d1dc9820eb4c4d3196f"
 dependencies = [
  "leb128fmt",
- "wasmparser 0.248.0",
+ "wasmparser 0.252.0",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


gbj:
See e2dbc0e8a7f4e03835d7d6522acda82adc1eda23

PR #667, opened 2026-07-04
feat: add the ability to pass the arguments `--locked`, `--offline`, and `--frozen` to cargo

### LINKED ISSUE #663: Build command does not support `--locked`
I want to build the application according to the `Cargo.lock` file without touching it but `cargo leptos build` does not support `--locked` and passing `--bin-cargo-args=--locked --lib-cargo-args=--locked` still updates the `Cargo.lock` file and thus is not suitable for deterministic builds like in CI pipelines. Is there anything I've overlooked or is this simply not supported (yet)?

ProbstDJakob:
Closes: #663


ProbstDJakob:
As the snapshots are not stable enough, should I disable them again or is there a way to make them reproducible?

gbj:
My understanding of the snapshot failures is that they are generating different file hashes on different runs and on different platforms, and that's what's failing. I don't have an answer to whether there's a way to make this deterministic but it seems unrelated to this PR, so I don't think it's a particularly useful test to add for this PR. What are you intending to test with those snapshots? Or can they just be removed?

ProbstDJakob:
They can be removed. The tests themselves should still persist, but the assertion with the snapshot could be removed/commented again.

ProbstDJakob:
Hmm, the test breaks because of `CARGO_TERM_COLOR=always`. I'll think about it to get this solved without "parsing" the ANSI escape codes to skip them or using unsafe to overwrite the env var. Maybe I'll just use a substring, but perhaps I find a better solution.

ProbstDJakob:
This should now work (at least in my local tests).

ProbstDJakob:
I do not have windows to test, but this should do the trick.

gbj (approved):
Looks good, thanks!

PR #671, opened 2026-07-21
fix: detect host musl libc at runtime instead of compile time

tcerqueira:
## Problem

`is_linux_musl_env()` used `cfg!(target_env = "musl")`, which reflects how *cargo-leptos itself* was compiled, not the host it runs on. A statically linked musl build reports musl on any Linux host, so on glibc systems it downloads musl-linked tool binaries that then fail to dynamically link.

Concretely, this breaks CI. `taiki-e/install-action` installs the musl build of cargo-leptos on Ubuntu runners; that binary runs fine (it's static), but `is_linux_musl_env()` returns `true`, so `cargo leptos build` fetches `tailwindcss-linux-x64-musl`, whose ELF interpreter `/lib/ld-musl-x86_64.so.1` doesn't exist on glibc Ubuntu. The build fails with an opaque:

```
Error:
   0: No such file or directory (os error 2)
Location: src/ext/sync.rs:83
```

## Fix

Detect the host libc at runtime via `ldd --version` (glibc prints to stdout, musl to stderr), cache the result, and fall back to the compile-time value when detection is inconclusive (e.g. `ldd` unavailable).

I did not test it, but the fallback provides the same functionality as before.

---

_Supersedes #669 — moved off my fork's `main` onto a dedicated `fix-musl-detection` branch._

### DIFF
--- a/src/ext/util.rs
+++ b/src/ext/util.rs
@@ -24,8 +24,38 @@ pub fn os_arch() -> Result<(&'static str, &'static str)> {
     Ok((target_os, target_arch))
 }
 
+/// Whether the *host* uses linux and musl libc. Best-effort detection at runtime, but
+/// falls back to compile time if it fails.
 pub fn is_linux_musl_env() -> bool {
-    cfg!(target_os = "linux") && cfg!(target_env = "musl")
+    #[cfg(not(target_os = "linux"))]
+    {
+        false
+    }
+    #[cfg(target_os = "linux")]
+    {
+        use std::sync::OnceLock;
+
+        static IS_MUSL: OnceLock<bool> = OnceLock::new();
+        *IS_MUSL.get_or_init(|| detect_host_libc_is_musl().unwrap_or(cfg!(target_env = "musl")))
+    }
+}
+
+#[cfg(target_os = "linux")]
+fn detect_host_libc_is_musl() -> Option<bool> {
+    use regex::Regex;
+    use std::process::Command;
+    use std::sync::LazyLock;
+
+    static LIBC_RE: LazyLock<Regex> =
+        LazyLock::new(|| Regex::new(r"(?i)(musl)|glibc|gnu libc|gnu c library").unwrap());
+
+    let detect = |stream: &[u8]| {
+        let text = String::from_utf8_lossy(stream);
+        LIBC_RE.captures(&text).map(|caps| caps.get(1).is_some()) // 'musl' capture group
+    };
+
+    let output = Command::new("ldd").arg("--version").output().ok()?;
+    detect(&output.stdout).or_else(|| detect(&output.stderr))
 }
 
 pub trait StrAdditions {


gbj:
> I did not test it

Could you clarify -- what exactly did you not test, and what did you test?

tcerqueira:
@gbj I tested on macos with the runtime environment the same as the build environment which is probably 99% of the cases and probably why this went unoticed.

I didn't thoroughly tested it because it's a bit of a pain and I would prefer some feedback before putting that effort. If you think the changes directionally make sense I'll properly test it.

gbj:
> I didn't thoroughly tested it because it's a bit of a pain and I would prefer some feedback before putting that effort. If you think the changes directionally make sense I'll properly test it.

I know almost nothing about the topic, unfortunately, so I'm not very qualified to say whether it makes sense directionally. 

I'm not sure I understood your reply about the testing. Is it that
A) you encountered this issue in your CI, you have proposed this fix, but you have not tested whether this fixes your CI; or
B) you encountered this issue in CI, have tried this fix, and it fixes your CI; or
C) you yourself have not encountered this issue, but are proposing this fix, so haven't set up a test environment for it; or
D) something else?

tcerqueira:
Closer to A) but I encountered this issue when I used the `taiki-e/install-action` for installing `cargo-leptos` on my CI. When I found the issue I just worked around the limitation and didn't use `taiki-e/install-action` action which pre-compiles with musl; instead I directly curl'ed the pre-compiled binary from github releases and installed it.

Basically my CI works either way right now, but people using the `taiki-e/install-action` action, which pre-compiles with musl, will hit this problem. I haven't directly tested this use case because I don't think I can, but I can set up some cross compilation containers and assert it downloads the correct tailwind binary.

### Need feedback

The fix for detecting is shelling out to `ldd --version` and parse the output. Is there a better way to do this?

gbj:
`ldd` seems like a reasonable approach. Another one might just be to check for the MUSL loader file?
```rust
std::path::Path::new(&format!("/lib/ld-musl-{arch}.so.1")).exists()
```

For testing... I'd imagine this can be tested locally with a Dockerfile?

But like I said I'm not very familiar with this topic, on either of those points.

tcerqueira:
So, I have a Dockerfile that builds `cargo-leptos` either on main or this ref (passed as arg), creates the most minimal app to reproduce the issue and then runs a build in both musl and glibc hosts. (Dockerfile at the end)

### 1. The bug, on `main`

```sh
docker build --build-arg REF=main -t cl:main - < Dockerfile
docker run --rm cl:main
```

```
ldd (Debian GLIBC 2.36-9+deb12u13) 2.36
...
     Install downloading tailwindcss https://github.com/tailwindlabs/tailwindcss/releases/download/v4.2.1/tailwindcss-linux-arm64-musl
     Command using tailwindcss v4.2.1 /root/.cache/cargo-leptos/tailwindcss-v4.2.1/tailwindcss-v4.2.1/tailwindcss-linux-arm64-musl
Error:
   0: No such file or directory (os error 2)

Location:
   src/ext/sync.rs:83
```

That is the reported CI failure verbatim. (`-x64-musl` on x86_64 machines.)

### 2. This PR

```sh
docker build --build-arg REF=pull/671/head -t cl:pr - < Dockerfile
docker run --rm cl:pr
```

```
ldd (Debian GLIBC 2.36-9+deb12u13) 2.36
...
     Install downloading tailwindcss https://github.com/tailwindlabs/tailwindcss/releases/download/v4.2.1/tailwindcss-linux-arm64
    Tailwind finished tailwindcss --input input.css --output /app/target/tmp/tailwind.css
```

Exit 0.

### 3. No regression on real musl hosts

Same static musl binary, Alpine instead of Debian:

```sh
docker build --build-arg REF=pull/671/head --target musl-host -t cl:pr-musl - < Dockerfile
docker run --rm cl:pr-musl
```

```
musl libc (aarch64)
...
     Install downloading tailwindcss https://github.com/tailwindlabs/tailwindcss/releases/download/v4.2.1/tailwindcss-linux-arm64-musl
    Tailwind finished tailwindcss --input input.css --output /app/target/tmp/tailwind.css
```

Exit 0, and it picks the same binary `main` picks there
(`docker build --build-arg REF=main --target musl-host ...` to confirm).

### Dockerfile 

```dockerfile
# syntax=docker/dockerfile:1.7
#
# `build`      compiles cargo-leptos into a *static musl* binary — exactly what
#              `taiki-e/install-action` installs on a CI runner.
# `app`        a minimal crate whose only cargo-leptos step is Tailwind.
# `glibc-host` runs that musl binary on Debian (glibc)  -> the bug.
# `musl-host`  runs that musl binary on Alpine (musl)   -> must keep working.

ARG REF=main

FROM rust:1-bookworm AS build
ARG REF
RUN apt-get update \
 && apt-get install -y --no-install-recommends musl-tools \
 && rm -rf /var/lib/apt/lists/*
WORKDIR /src
RUN git init -q . && git remote add origin https://github.com/leptos-rs/cargo-leptos.git
RUN git fetch -q --depth 1 origin "$REF" && git checkout -q FETCH_HEAD && git log --oneline -1
# after the checkout, so the target lands on the toolchain rust-toolchain.toml selects
RUN rustup target add "$(uname -m)-unknown-linux-musl"
RUN --mount=type=cache,target=/usr/local/cargo/registry,sharing=locked \
    --mount=type=cache,target=/src/target,sharing=locked \
    T="$(uname -m)-unknown-linux-musl"; \
    cargo build --target "$T" && cp "target/$T/debug/cargo-leptos" /cargo-leptos

FROM rust:1-bookworm AS app
WORKDIR /app
RUN <<'EOF'
set -eu
mkdir -p src
cat > Cargo.toml <<'TOML'
[package]
name = "minimal"
version = "0.1.0"
edition = "2021"

[lib]
crate-type = ["cdylib", "rlib"]

[dependencies]
wasm-bindgen = "=0.2.104"

[package.metadata.leptos]
tailwind-input-file = "input.css"
TOML
cat > src/lib.rs <<'RS'
use wasm_bindgen::prelude::*;

#[wasm_bindgen]
pub fn hello() -> String {
    "hi".to_string()
}
RS
echo 'fn main() {}' > src/main.rs
echo '@import "tailwindcss";' > input.css
EOF

# musl cargo-leptos on a musl host: must still pick the -musl tool binaries
FROM rust:1-alpine AS musl-host
RUN rustup target add wasm32-unknown-unknown
COPY --from=build /cargo-leptos /usr/local/bin/cargo-leptos
COPY --from=app /app /app
WORKDIR /app
RUN cargo build --lib --target wasm32-unknown-unknown --target-dir=target/front
CMD ["sh", "-c", "(ldd --version 2>&1 || true) | head -1; cargo-leptos build --frontend-only -v"]

# musl cargo-leptos on a glibc host: this is the bug
FROM rust:1-bookworm AS glibc-host
RUN rustup target add wasm32-unknown-unknown
COPY --from=build /cargo-leptos /usr/local/bin/cargo-leptos
COPY --from=app /app /app
WORKDIR /app
RUN cargo build --lib --target wasm32-unknown-unknown --target-dir=target/front
CMD ["sh", "-c", "(ldd --version 2>&1 || true) | head -1; cargo-leptos build --frontend-only -v"]
```

gbj (approved):
Beautiful! Thanks so much for testing it, and I think this sounds like it's good to go.

PR #672, opened 2026-07-24
fix(watch): honor --frontend-only/--server-only on rebuilds

### LINKED ISSUE #670: `--frontend-only` is ignored by `cargo leptos watch` on rebuilds
## Summary

`cargo leptos watch` accepts `--frontend-only` and honors it for the *initial* build, but every subsequent rebuild triggered by a file change builds the server anyway. The flag silently stops applying after the first pass.

Version: 0.3.7 (verified against the `v0.3.7` tag source; the relevant files are identical on `main`).

## Detail

`--frontend-only` is defined on the shared `Opts` (`src/config/cli.rs:73-75`), and `Watch(BinOpts)` wraps `Opts` (`src/config/cli.rs:204`, `:166`), so the flag parses fine under `watch` and lands in `proj.build_frontend_only` (`src/config/project.rs:170`).

`watch()` starts with `build_proj(proj)` (`src/command/watch.rs:17`), which does respect it (`src/command/build.rs:65`: `let needs_server = !proj.build_frontend_only;`).

But the rebuild path does not. `run_loop` → `runner()` calls `compile::server` unconditionally:

```rust
// src/command/watch.rs:98
let server_hdl = compile::server(proj, &changes).await;
let front_hdl = compile::front(proj, &changes).await;
```

`grep -rn 'build_frontend_only\|build_server_only' src/` returns only:

```
src/command/build.rs:64:    let needs_frontend = !proj.build_server_only;
src/command/build.rs:65:    let needs_server = !proj.build_frontend_only;
src/config/project.rs:71,72,170,171
```

Nothing in `src/command/watch.rs` reads either field, so `--server-only` is presumably inconsistent under `watch` in the same way.

## Reproduce

```
cargo leptos watch --frontend-only
```

Touch a file under the lib source paths. The initial build skips the server; the rebuild does not.

## Expected

Either `runner()` gates the server compile on `proj.build_frontend_only` the way `build_proj` does, or `watch` rejects `--frontend-only`/`--server-only` up front rather than accepting a flag it only partly applies.

I'm happy to open a PR for the first option if that's the preferred direction.

pat-s:
## Summary

`build_proj` gates the initial build on `--frontend-only` / `--server-only`, but `runner` (the rebuild path in `watch`) compiled the server, front, assets and style unconditionally.
So `cargo leptos watch --frontend-only` (and `--server-only`) only applied to the first build and was silently ignored on every file-change rebuild.

This gates the rebuild compiles on the same `build_frontend_only` / `build_server_only` flags, so the options apply uniformly to the initial build and all subsequent rebuilds.

Closes #670

### DIFF
--- a/src/command/watch.rs
+++ b/src/command/watch.rs
@@ -8,9 +8,9 @@ use crate::{
 };
 use leptos_hot_reload::ViewMacros;
 use std::sync::Arc;
+use tokio::join;
 use tokio::sync::broadcast::error::RecvError;
 use tokio::task::JoinHandle;
-use tokio::{join, try_join};
 
 pub async fn watch(proj: &Arc<Project>) -> Result<()> {
     // even if the build fails, we continue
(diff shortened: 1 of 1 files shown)


gbj (approved):
This looks good, thanks!

Issue #586, opened 2025-11-07
Watch does not track imported CSS files


fogzot:
When using Tailwindcss it is common to split the CSS over multiple files (for example we have `colors.css`, `tokens.css`, etc.) Curently `cargo leptop watch` regenrates the CSS only when the main CSS file changes and completely ignore '@import`'ed CSS files.

I tried to use the `watch-additional-files` option but I did not found a way to force a reload when one of the imported files changes.

It would be nice if `cargo leptos watch` could track all CSS files when building using Tailwind.

Issue #444, opened 2025-02-20
hot-reload not working for `tailwind-input-file`


1xX69:
### Steps to reproduce:

1. Install latest `cargo-leptos` from GitHub (bleeding-edge version).
2. Set `export LEPTOS_TAILWIND_VERSION="v4.0.6"`
3. Set `tailwind-input-file` to a CSS file (e.g. `tailwind.css`).
4. Run `cargo leptos watch --hot-reload -vv`.
5. Make changes to `tailwind.css`.

**Issue**: Changes are detected but not applied until manual restart.

**Expected**: Hot reload should update CSS changes automatically.

**Context**: This feature is crucial for building classes with Tailwind's [@apply](https://tailwindcss.com/docs/functions-and-directives#apply-directive) directive, which relies processing of `tailwind-input-file`. Without hot reload, development is hindered.

1xX69:
Note that hot reload works for `style-file`, but it's not a suitable workaround since PostCSS processing (and therefore Tailwind's @apply feature) only applies to `tailwind-input-file`.

benwis:
The biggest issue I see here is not Leptos related, it's that Tailwind itself will only include css classes in the code. in it's output. There's just about nothing we can do here.

Greg gave a demo showing off tailwind CSS hot reload yesterday, but he did it by including the tailwind css from their CDN. 

In short, if you're adding or removing an existing CSS class or editing the static part of the view! macro, that will work with hot-reload. Tailwind as it's described won't.

I'm going to close this, but if someone has an idea as to how we could change this, feel free to reopen or submit a PR

ghost:
Using [@apply](https://tailwindcss.com/docs/functions-and-directives#apply-directive) directive, one can develop CSS-only components like buttons, images and etc. Shouldn't this improve compilation times for development builds cuz the components are pure CSS?

ghost:
Shouldn't this issue remain open until the problem is solved?

gbj:
@1xX69 Any chance you are interested in working on a PR?

ghost:
> [@1xX69](https://github.com/1xX69) Any chance you are interested in working on a PR?

I'm not the best person for this, but thank you for thinking of me. Who else is interested in giving it a try? Just leave a comment below and gbj might assign you!

P.s. I've added current workaround to the description.

PR #674, opened 2026-08-03
feat: Add support for bundling CSS files

### LINKED ISSUE #238: When using a plain css style sheet, lightning should be configured to bundle in any @import style sheets
Given the following:

```css
@import "../generated/leptonic/leptonic-themes";

body {
	font-family: sans-serif;
	text-align: center;
}

:root {
	--brand-color: green;
}
```

If it is named as main.scss then sass resolves the import and it gets bundled as one file
`style-file = "style/main.scss"`

If it is named as main.css, then the import is not bundled by the current lightning config. 
`style-file = "style/main.css"`

spencewenski:
Currently, Lightning CSS is only used to transpile and minify a single CSS file (either literally a single `.css` file, or a CSS file generated by Sass or Tailwind). However, Lightning CSS also supports bundling plain CSS files that use the native CSS `@import` at-rule.

This PR adds support for bundling the `style-file` and any files it imports into a single CSS file. This allows using plain CSS in a project while still getting the benefit of splitting CSS into separate smaller files instead of maintaining a single large CSS file.

The bundling functionality is triggered if the new `style-bundle` configuration is set to `true` and the `style-file` has the `.css` file extension.

Also, now that CSS files can be bundled, it may be useful to co-locate them next their corresponding source file, e.g., placing `button.css` and `button.rs` in the same directory. The files can then be watched by adding a directory containing all bundled css files (e.g. `src/`) to `watch-additional-files`. However, I noticed that doing so results in rebuilding the frontend instead of rebuilding the styles:

```
Notify handle ["src/colocated_css_file.css"]
Notify additional file change src/colocated_css_file.css
Style no build needed ChangeSet([Additional])
Build step done with changes: Front
Watch updated Front
```

This PR updates the `watch_additional_files` config and handling to allow more control over how the additional file changes are handled:

```
Notify handle ["src/colocated_css_file.css"]
Notify Style change src/colocated_css_file.css
Front no changes to rebuild
Style found: style/main.css -> @pkg/css_bundle_example.css
Style finished with changes @pkg/css_bundle_example.css
Build step done with changes:
Watch updated style
```

- Resolves #238
- I _think_ this will also resolve:
    - #586
    - #444


src/service/notify.rs
```diff
@@ -206,11 +207,16 @@ fn handle(
...
 
         if path.starts_with_any(&proj.watch_additional_files) {
-            debug!(
-                "Notify additional file change {}",
-                GRAY.paint(path.as_str())
-            );
-            changes.add(Change::Additional);
+            if path.is_ext_any(&STYLE_EXTENSIONS) {
+                debug!("Notify style change {}", GRAY.paint(path.as_str()));
+                changes.add(Change::Style);
```
gbj: I think this changes some existing, valid behavior, by defining CSS file changes as *only* style changes. There are a few different ways that CSS file changes could be more than that; for example, if you `include_str!("some_file.css")` then it needs to rebuild the binary. 

Not sure how best to handle that. A separate flag for "watch these files but only as CSS?" Gating this change on the bundling feature being activated?

spencewenski: Good point, I hadn't considered that. I think I came up with a decent solution. I expanded the `watch-additional-files` config to allow providing either paths or structures, where the structure allows configuring the change types to trigger when a file is modified.

I also added a new example package including a `Cargo.lock` file, which is why the git changeset looks much bigger now

spencewenski:
@gbj , not sure if you saw the update to this PR -- I believe it should be ready to re-review. No rush, just wanted to reach out in case you missed the updates.

gbj:
This looks good except that I think the new docs in the README are wrong -- they list `path = `, `extension = `, and `change-type = ` in the example but the fields are named `paths`, `extensions`, and `change_types` (→ `change-types`). Shouldn't they be plural in the README?

spencewenski:
Oops, yes, I originally had them as singular/non-Vec fields but then I decided to make them plural/Vec fields and forgot to update the names in the readme. Just pushed the fix.

gbj (approved):
Thanks! Looks good to me.

PR #675, opened 2026-08-06
Remove the re-patching of the __wasm_split file

chrisgrounds:
**Background**

We were deploying our leptos app with the `--split` feature and randomly found it to break upon not being able to find the js entrypoint file. The name of the file was clearly malformed, eg: `website.jsXXX.jsXXX.js`.

**The Problem**

My understanding of what was happening here is as follows: the `__wasm_split` file is the loader file which references the wasm chunk files as well as the hashed main JS entrypoint. We already `replace_in_file` on line 21 of `hash.rs` but for the `__wasm_split` file were doing it again in the `replace_wasm_split_references` function.

With the code calling `replace_in_file` twice most of the time it was a no-op and an idemopotent operation. However, if the hash ever started with "js", then `replace_in_file` would incorrectly insert the hash a second time, treating "js" as an extension and not part of the hash.

For example,

| Phase | Output |
| --- | ----------- |
| original filename      | `website.js`       |
| hash   | `jsXXX`        |
| first `replace_in_file` pass (i.e. replace original filename with hashed version)   | `website.jsXXX.js`       |
| second `replace_in_file` pass (because it matches on original filename which incorrectly includes part of the hash and replaces it again)    | `website.jsXXX.jsXXX.js`       |

This would then cause the loading of the file to fail as that filename would be incorrect.

I think the risk of this problem arising is 1 in 4096 as it requires the first characters of the hash to be "js".

**Solution**

My proposed solution is to remove the branch in `replace_wasm_split_references` that was running the `replace_in_file` on `__wasm_split`. 

I was thinking about keeping the branch and adding a condition to ensure the loader file was not reprocessed,

```rust
else if filename.starts_with("__wasm_split") && filename != new_wasm_split_filename {
    replace_in_file(
      &Utf8PathBuf::try_from(path).unwrap(),
      renamed_files,
      pkg_dir,
      false,
    );
}
```

But, as far as I can see, that just means the branch will never be executed at all. So I've decided to just delete the branch instead.

Not sure if this is the best solution or introduces latent risk. Happy to discuss or if someone has a better idea in mind happy to change to that!

**Tests**

Added two tests to cover the bug. The first test checks I've not introduced a regression. The second checks that the bug is resolved. If you re-add the branch that I've deleted then the test will fail with the following, verifying the existence of the bug:

```
   thread 'compile::hash::tests::when_hash_starts_with_extension_does_not_repatch_the_wasm_split_loader_it_already_patched' (9454187) panicked at src/compile/hash.rs:399:9:
    assertion `left == right` failed: loader file was patched a second time, corrupting its own reference
      left: "import { initSync } from \"/pkg/website.jsABCDEFGHIJKLMNOPQRSTU.jsABCDEFGHIJKLMNOPQRSTU.js\";"
     right: "import { initSync } from \"/pkg/website.jsABCDEFGHIJKLMNOPQRSTU.js\";"
```

**AI Disclaimer**

We used Sonnet 5 to help investigate and debug, as we initially thought the problem lay with our code. Sonnet soon started investigating the hashing functionality in `cargo-leptos` and identified the issue here. The code deletion was performed by a human and not an LLM.


gbj (approved):
Looks good, thanks!

PR #678, opened 2026-08-29
Skip front split/bindgen when the emitted wasm is unchanged

svieujot:
## Problem

In watch mode the front task's change detection is feature-agnostic: `src_paths` come from cargo-metadata path dependencies without feature resolution, so a change that only affects the server binary (or a `watch-additional-files` entry) still triggers the front build. Cargo then finishes as a near no-op — but splitting (`--split`) and wasm-bindgen run unconditionally afterwards, reprocessing a byte-identical module.

On a large workspace this dominates the rebuild: with a ~1.1 GB emitted wasm module and `--split`, the split + wasm-bindgen stages cost 40–120 s per rebuild depending on the machine and toolchain build, and they run again on every server-only edit, every locale tweak, every watched-asset change.

## Change

After `cargo build` finishes, hash the emitted wasm artifact and compare it with the input of the last successful split/bindgen run in this process:

- unchanged → skip splitting and wasm-bindgen, log the usual completion line, and return `Product::None` — the server still restarts immediately and no full browser reload is sent;
- when the trigger includes a watched additional file, the skip returns `Product::Assets` instead, keeping the browser reload such files rely on;
- the hash is recorded only after a *successful* pipeline run, so a failed or interrupted run is retried on the next cycle;
- the first build in a process always runs (the site directory is repopulated at startup), and so does any read failure or a missing bindgen output.

## Measured effect

On our project (1.5m lines), a server-only edit went from ~2–4 minutes (full split + bindgen) to the server compile + restart alone (~20 s); CSS/asset-only cycles complete in ~2 s. Wasm-changing edits are unaffected (one extra ~0.3 s hash before the pipeline).


gbj:
Thanks for the PR! This is definitely a good idea.

Two small changes I'd suggest:
1. This should probably remove the old hash for the file before the process of building the file starts, rather than just overwriting it on success; otherwise, errors or anything else that interrupts the build process will leave a stale hash in the table
2. With the PR, if you have wasm-splitting enabled, `front_wasm_hash()` will read the whole file at line 57 (in `front.rs`), hash it, and drop it. Inside the `if proj.split` block at line 71, we'll then re-read the whole file again (line 75), split it, and drop it again. For the 1.1gb file case, that means reading the same 1.1gb file into memory twice, with the time cost of reading it/allocating the memory/deallocating the memory. If you hoist the `let input_wasm = tokio::fs::read(&proj.lib.wasm_file.source).await?;` line up to line 56, and pass the `input_wasm` into you can read the file once, hash it, split it as necessary, then drop it once. That should make your version even faster.

```diff
diff --git a/src/compile/front.rs b/src/compile/front.rs
index e85ea09..5381328 100644
--- a/src/compile/front.rs
+++ b/src/compile/front.rs
@@ -54,9 +54,12 @@ pub async fn front(
         debug!("Cargo envs: {}", GRAY.paint(envs));
         info!("Cargo finished {}", GRAY.paint(line));
 
-        let wasm_hash = front_wasm_hash(&proj).await;
+        let input_wasm = tokio::fs::read(&proj.lib.wasm_file.source).await?;
+        let wasm_hash = Some(seahash::hash(&input_wasm));
         if front_wasm_unchanged(&proj, wasm_hash) {
             info!("Finished generating JS/WASM for front (wasm unchanged; reusing previous output)");
             // A watched additional file can influence the running app without
             // changing the wasm (that is why it is watched): keep the browser
             // reload the full pipeline used to cause for those changes.
@@ -72,8 +75,6 @@ pub async fn front(
             info!("Front splitting out lazy-loaded WASM files");
             let start_time = tokio::time::Instant::now();
 
-            let input_wasm = tokio::fs::read(&proj.lib.wasm_file.source).await?;
-
             let split_files = wasm_split_tools::wasm_split(&input_wasm, false, &proj).await?;
             files.extend(split_files);
 
@@ -90,18 +91,6 @@ pub async fn front(
     })
 }
 
-/// Content hash of the cargo-emitted front wasm artifact, observed between
-/// the cargo build and the split/bindgen pipeline. `None` when the artifact
-/// cannot be read. The artifact was just written by cargo, so the read is
-/// served from the page cache and costs a fraction of a second even on
-/// gigabyte modules.
-async fn front_wasm_hash(proj: &Project) -> Option<u64> {
-    let data = tokio::fs::read(proj.lib.wasm_file.source.as_std_path())
-        .await
-        .ok()?;
-    Some(seahash::hash(&data))
-}
-
 fn front_wasm_registry() -> &'static Mutex<HashMap<String, u64>> {
     static REGISTRY: OnceLock<Mutex<HashMap<String, u64>>> = OnceLock::new();
     REGISTRY.get_or_init(Default::default
```

svieujot:
Thank you. I did those changes with small changes.

 I placed the take right after cargo finishes rather than at the top of the task: only split and bindgen write into the site dir, while cargo only writes the source .wasm, so a failed or interrupted cargo build leaves the recorded output genuinely valid. Taking before cargo would also drop the entry on every cargo failure, meaning the first good build after a typo pays the full pipeline for an unchanged wasm.

gbj (approved):
Looks good, thanks!

PR #681, opened 2026-09-15
fix: rewrite `__wasm_split_manifest.json` per whole chunk name

### LINKED ISSUE #680: `--split` + `hash-files` rewrite manifest chunk names that are prefixes of other chunk names
With `cargo leptos build --release --split` and `hash-files = true`, `__wasm_split_manifest.json` contains chunk names that don't exist on disk, so the `<link rel="preload">` tags leptos emits from it 404:

```
GET /pkg/chunk_1824.UNQnOA7-hefUE_sCbwvavQ8.UNQnOA7-hefUE_sCbwvavQ.wasm 404
GET /pkg/chunk_1825.UNQnOA7-hefUE_sCbwvavQ3.wasm 404
```

The site still works because `__wasm_split.<hash>.js` is rewritten correctly; only the preloads are broken.

adrianncovaci:
`replace_in_file` rewrote the manifest with a bare `contents.replace("chunk_1", "chunk_1.<hash>")`, so every chunk whose number is a prefix of another (`chunk_1` / `chunk_10`, `chunk_1824` / `chunk_18248`, ...) also corrupted those entries.

fix #680


gbj (approved):
Oof (to the bug) and thanks (for the PR). This looks good to me and I'll cut a patch release shortly to include it.

